PR #15, opened 2020-09-14
Subscribe to newsletter at the bottom of each blog post

dtuite:
The goal here is to ensure that people who put their emails in the form at the bottom of blog posts end up on a newsletter subscriber list where they will receive the weekly newsletter.

The form submissions are handled by [Netlify's forms feature](https://docs.netlify.com/forms/setup/). Each submission is sent to [Zapier](https://zapier.com), where it is forwarded on to the main subscriber list in [Revue](https://getrevue.co). Doing it this way allows us to keep the Revue API key out of the website out of people's browsers.

When the form is submitted successfully, a modal appears to congratulate them.

Landing page submissions go to a different list than those received from the bottom of blog posts. Those people don't really intend to sign up for a newsletter. We can choose to manually add them to the newsletter if we later get their email consent. I expect we will replace the landing page CTA with a "Get a demo" type flow fairly soon so I thought it wasn't worth messing with this.

### DIFF
--- a/src/components/actions/FormSubmissionModal.js
+++ b/src/components/actions/FormSubmissionModal.js
@@ -50,7 +50,7 @@ const FormSubmissionModal = ({
       <div className={classes.modalContentWrapper}>
         <h2>
           {titleText}{' '}
-          <span aria-label="Party Steamers" role="img">
+          <span aria-label="Party Streamers" role="img">
             🎉
           </span>
         </h2>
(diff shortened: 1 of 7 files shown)


iain-b (approved):

src/components/actions/FormSubmissionModal.js
```diff
@@ -44,15 +49,12 @@ const FormSubmissionModal = ({ modalOpen, handleCloseModal }) => {
     >
       <div className={classes.modalContentWrapper}>
         <h2>
-          Thank you!{' '}
+          {titleText}{' '}
           <span aria-label="Party Steamers" role="img">
```
iain-b, line 53: Typo ?  Streamers  with two "r"s?

dtuite: Good spot. Fix incoming.

PR #17, opened 2020-09-22
Add content from newsletter to blog

dtuite:
Add the content from the first two newsletter editions for better discoverability.

The styling isn't great on these pages. For example, HTML quotes are not highlighted. Improving the design is a larger effort though and I hope we can solve that separately to this PR.


src/templates/BlogPost.js
```diff
@@ -59,7 +59,11 @@ const BlogPostTemplate = ({ data, location }) => {
         <div className={classes.callToActionWrapper}>
           <InterstitialTitle text="Become a Backstage expert" />
           <p className={classes.callToActionParagraph}>
-            Get the latest Backstage news in your inbox each week.
+            To get the latest news, deep dives into Backstage features, and a roundup of recent
+            open-source action, sign up for Roadie&apos;s Backstage Weekly.{' '}
```
martina-if, line 63: out of curiosity, what is `{' '}` for?

martina-if (approved):

src/templates/BlogPost.js
dtuite: It's to add a whitespace between the `Weekly.` and the `<a href...`. If I remove it, HTML collapses the whitespace and the text of the anchor tag pushes up against the period.

The other option I know is to add a `&nbsp;` but that will refuse to line break correctly so AFAIK this solution is the best way to do it.

PR #19, opened 2020-09-25
Register for demo

dtuite:
Users can now insert their email to get a demo of Backstage.

 - [ ] Update the Zapier email template which is sent to people who sign up.
 - [x] Test the newsletter subscription flow on Netlify.
 - [x] Make sure the form at the bottom of the plugins is for newsletter sign up list. Not "Notify me".
 - [x] Set up a Zap to automatically subscribe certain emails to the newsletter on Revue.
 - [x] Add a newsletter link to the header.


src/components/SitewideFooter.js
```diff
@@ -32,14 +32,21 @@ const useStyles = createUseStyles((theme) => ({
...
     <footer>
       <div className={classes.inner}>
         <div>© {new Date().getFullYear()} Larder, Inc. All rights reserved.</div>
 
         <nav>
-          <Link to="/terms">Terms of service</Link>
+          <a href={newsletterUrl} target="_blank" rel="noopener noreferrer">
+            Backstage Weekly
+          </a>
+          <Link to="/terms" className={classes.nonFirstLink}>
```
martina-if, line 47: out of curiosity what is  `classes.nonFirstLink`?

src/components/SitewideHeader/HamburgerMenu.js
```diff
@@ -96,9 +96,12 @@ const useStyles = createUseStyles(() => ({
   },
 }));
 
-const HamburgerMenu = () => {
+const HamburgerMenu = ({ siteMetadata }) => {
   const classes = useStyles();
 
+  const TWITTER_URL = `https://twitter.com/${siteMetadata.social.twitter}`;
```
martina-if, line 102: Why not store the whole url in the siteMetadata instead?

martina-if:
The new emoji is also missing for me:

demo-emoji-missing

If I sign up for the demo, I'm not supposed to sign up for the newsletter? Or maybe I had already done that before :/
 
already

martina-if (dismissed):

dtuite:
@martina-if 

Signing up for the demo doesn't automatically sign people up for the newsletter. Doing so would be a GDPR violation unless we explicitly state that the user is going to be signed up to a newsletter. We could add a checkbox under the form which says "sign me up for the newsletter" but that feels like too much clutter for a landing page. I have been manually checking with people if they would like to sign up for the newsletter. It's a manageable solution for now.

I don't understand the issue with the emojis. You use Linux right? What browser are you using? Are they also missing on the production deployment of roadie.io?

src/components/SitewideFooter.js
dtuite: I'm adding spacing between the links to keep them apart from each other but the spacing is not required on the first link in the list.

src/components/SitewideHeader/HamburgerMenu.js
dtuite: This setup (twitter username but not URL in config) comes with Gatsby by default so I kinda just stuck with it. I guess one advantage is that we have the twitter username available if we ever need that in non-linked form. We can count on the Twitter hostname not changing anytime soon so it's basically static config.

With that said, I'm not actually using the metadata version in the code where I reference the username. I will fix that.

martina-if:
Ok, I think I got that message "The email address has already been subscribed" because I got redirected to the production page instead of the netlify preview. I thought I couldn't sign up for both independently but it's not the case :+1: all good.

martina-if (approved):

PR #21, opened 2020-09-30
Update travis-ci docs

mcalus3:


content/plugins/descriptions/travis-ci.yaml
```diff
@@ -19,24 +19,12 @@ coverImage:
   alt: 'A screenshot of the Travis CI plugin. It is showing a list of builds for a sample service.'
 
 gettingStarted:
-  - intro: 'Install the plugin in your Backstage instance'
```
iain-b, line 22: Users who create their own app rather than using the sample app may have to do this; AFAIK this plugin isn't part of the create app template? It's worth listing all the steps which a user may need to follow and then suggesting they can skip some steps in certain circumstances (i.e. they're running the example app)

content/plugins/descriptions/travis-ci.yaml
```diff
@@ -19,24 +19,12 @@ coverImage:
...
-      export { plugin as TravisCI } from '@roadiehq/backstage-plugin-travis-ci';
-  - intro: 'Add the API to your Backstage instance'
-    language: 'typescript'
-    code: |
-      // packages/app/src/api.ts
-      import { TravisCIApi, travisCIApiRef } from '@roadiehq/backstage-plugin-travis-ci';
-
-      // ...
-
-      builder.add(travisCIApiRef, new TravisCIApi());
```
iain-b, line 38: We should update this to be consistent with new way of adding plugins (plugins.ts)

content/plugins/notes/travis-ci.md
```diff
@@ -2,25 +2,17 @@
...
-To use the Travis CI Backstage plugin, first visit Travis CI to get an API token.
+## Features
 
-There are two versions of Travis CI: https://travis-ci.com and https://travis-ci.org. Tokens
-from .com do not work for .org and vice versa, so make sure you are on the right Travis website.
+- List Travis CI Builds
+- Retrigger builds
 
-Tokens can be copied from the settings page in Travis CI. Use the menu in the top right of the page to navigate to Settings. Click Settings again, under your name this time, and you should see API authentication.
+To use the Travis CI Backstage plugin, first visit Travis CI to get an API token: https://travis-ci.com/account/preferences
```
iain-b, line 10: I know we're pitching at a technical audience but it's worth being very specific about how they issue a token. A link to Travis CI docs is fine if there is one or you can write a few lines. The user will ignore it if they know what they're doing.

iain-b (changes requested):

content/plugins/notes/travis-ci.md
mcalus3: This link points to the settings page shown in the image, on which there is a first form field "API KEY" which has to be copied. This assumes that the user has already created an account.

PR #27, opened 2020-10-09
feat - add new code insights plugin

lewtakm:
Add new plugin description.


content/plugins/descriptions/code-insigts.yaml
```diff
@@ -0,0 +1,92 @@
+# The server must be restarted to pick up changes in ths file.
```
iain-b, line 1: Typo in the file name

iain-b (changes requested):
Typo in the filename but otherwise lgtm

content/plugins/descriptions/code-insigts.yaml
lewtakm: done

iain-b (approved):

PR #29, opened 2020-10-11
Add catalog-info.yaml

dtuite:
So we can have an example of a website in our Backstage demos.


dtuite:
Closing this. Branch names with slashes in them cannot be imported into Backstage. You get a 404 error when attempting to add them to a catalog. I've opened a new branch called `backstage-metadata` on this repo with the same diff. I will open a new PR from that branch.

I'm fairly sure an issue for this already exists on the Backsage upstream repo.

PR #30, opened 2020-10-11
Add catalog-info.yaml

dtuite:
So we can use a website in our Backstage demos.


dtuite:
Just realised there already is a metadata file. 😭

PR #37, opened 2020-10-15
Update catalog-info.yaml

iain-b:
I think this will mean ownership will work for one of us.


dtuite (approved):
If this works we should assign the different sample service catalog-info files to each of us.

PR #39, opened 2020-10-16
feat: add lambda plugin

lewtakm:
Fresh PR to "clean" commit history thats add Lambda plugin to Roadie website.


content/plugins/notes/aws-lambda.md
```diff
@@ -0,0 +1,17 @@
+---
+name: aws-lambda
+---
+
+## Authentication
+
+In order to perform requests to AWS lambda plugin first asks backend for temporary credentials via /api/aws/credentials.
+(it uses **@roadiehq/backstage-plugin-aws-auth** backend plugin)
```
dtuite, line 8: This should be a link to the GitHub of the plugin.

Does this auth plugin come installed in Backstage? If not, we need to clearly state that it needs to be installed.

dtuite (changes requested):

content/plugins/notes/aws-lambda.md
lewtakm: Sure I will add a link here. 

Sadly no, backend plugin is not installed in Backstage. I made lots of changes in Lambda plugin today because it is not working anymore at least not for new people. Backstage team changed annotation system and right now it is not possible to add annotations that was required by this plugin. Also plugin might break whole UI when installed without backend plugin, I added ErrorBoundary component to prevent it and show information thats backend plugin is required.

dtuite (approved):
Thanks for making those changes. 👍

PR #40, opened 2020-10-17
Point demo link to demo.roadie.so

dtuite:
Move the demo over to https://demo.roadie.so because it's deployed with the 1-click Kubernetes charts.

 - [ ] 301 redirect backstage-demo.roadie.io over to demo.roadie.so

### DIFF
--- a/src/templates/BlogPost.js
+++ b/src/templates/BlogPost.js
@@ -47,6 +47,7 @@ const BlogPostTemplate = ({ data, location }) => {
         handleCloseModal={handleCloseModal}
         titleText="You're subscribed!"
         bodyText="You should receive the first edition within a week."
+        siteMetadata={data.site.siteMetadata}
       />
 
       <StickyFooter maxWidthBreakpoint={MAX_WIDTH_BREAKPOINT} location={location}>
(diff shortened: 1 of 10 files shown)


src/templates/BlogPost.js
```diff
@@ -80,6 +81,10 @@ export const pageQuery = graphql`
     site {
       siteMetadata {
         title
+        demoUrl
```
martina-if, line 84: where does this info come from?

martina-if (approved):

martina-if:
Unrelated to this but There are a couple of icons/emojis not loading for me on the careers page:

careers

src/templates/BlogPost.js
dtuite: https://github.com/RoadieHQ/marketing-site/blob/master/gatsby-config.js#L7

dtuite:
@martina-if Even on [the production site](https://roadie.io/careers)? 

This is what I see in Chrome and Firefox.

Screenshot 2020-10-19 at 08 49 47

src/templates/BlogPost.js
martina-if: thanks

PR #45, opened 2020-10-28
Fix: remove highlight from operators in code preview

lewtakm:
This PR removes highlights from operator in the code preview:

Before:
o1JSz9M

After:
2pfwGaX


martina-if (approved):
nice! that was easy :grinning:

PR #46, opened 2020-10-30
feat: add plugins descriptions

lewtakm:
This PR adds plugins description for rest of the plugins from available in Backstage.

I also added one fix that include Pull Request plugin screenshot update.


lewtakm:
Not sure why but  all images was replaced by the last ones (lighthouse) on the preview. Anyone faced similar issue?

dtuite:
@lewtakm I didn't review in detail yet but it's awesome to see progress on this. Thank you so much.

Those last 3 plugins are worth finishing out. We've already had potential customers ask us about both NewRelic and the TechRadar plugin for example.

dtuite:
@lewtakm I have not seen that issue before no.

lewtakm:
> @lewtakm I have not seen that issue before no.

@dtuite  I figured out why that was happening. Something is wrong with the file system. For example if two images has same name e.x. logo.png but lay in different folders there are still threated as one. For now I just updated file names but this bug should be fixed. Should I create a Clubhouse task for this?

content/plugins/descriptions/api-docs.yaml
```diff
@@ -0,0 +1,65 @@
...
+  title: 'Backstage AWS Lambda Plugin | Roadie'
+  description: |
+    Components to discover and display API entities as an extension to the catalog plugin.
+
+logo:
+  # This must be a relative path. It should start without a slash or with ./
+  # The overall size of the file should be approximately 200 by 200.
+  # The image in the file should be approximately 100 by 100 pixels. It needs space around it.
+  # The image will be made greyscale by Gatsby Image Sharp.
+  fileSystemPath: './content/assets/logos/api-docs/logo-docs.png'
```
dtuite, line 22: The grey color in this logo is clashing with the green background.

Screenshot 2020-11-04 at 12 28 19

Is it possible to edit the logo image to make the grey section transparent?

content/plugins/descriptions/aws-lambda.yaml
```diff
@@ -19,7 +19,7 @@ logo:
   # The overall size of the file should be approximately 200 by 200.
   # The image in the file should be approximately 100 by 100 pixels. It needs space around it.
   # The image will be made greyscale by Gatsby Image Sharp.
-  fileSystemPath: './content/assets/logos/aws-lambda/logo.png'
+  fileSystemPath: './content/assets/logos/aws-lambda/logo-lambda.png'
```
dtuite, line 22: Same comment as the API Docs. 

Screenshot 2020-11-04 at 12 30 34

As a general rule, white and transparent should be the only colors used in logos.

If that's not viable, I'm open to changing the design of this page completely so that colored logos work better.

content/plugins/notes/cost-insights.md
```diff
@@ -0,0 +1,74 @@
+---
+name: cost-insights
+---
+
+## Cost Insights
```
dtuite, line 5: Copying content from other websites [can cause SEO problems and should be avoided](https://moz.com/learn/seo/duplicate-content). The content in question is open source, so from a licensing perspective it's fine to do this, but it confuses Google and won't help us to show up on search results.

I would prefer to see a tutorial which shows a user how to use the cost insights plugin, or a short, original, paragraph how to configure it. If you don't feel confident writing something like that, it is better to omit this section entirely.

Please replace this with something original or remove it entirely.

dtuite (changes requested):
Will reach out on Slack.

content/plugins/descriptions/aws-lambda.yaml
lewtakm: I made a preview how it might looks with colored logos: 
localhost_8000_backstage_plugins (1)

I my opinion it looks better and would be easier to maintain. I also changed sizes of the logos to be more compact. Also I removed visible white pixels from Firebase Functions logo.

lewtakm:
I updated the plugin descriptions. Now they are all unique.

Plugin page preview: 
localhost_8000_backstage_plugins (3)

dtuite (approved):
@lewtakm I reverted to a clean white look for the plugins directory.

Buildkite has a lowercase "K" by the way. Some changes might be required in the Buildkite code?

Screenshot 2020-11-19 at 14 54 32

PR #54, opened 2020-11-19
Revert path changes in logos

dtuite:
I changed some logos as part of a fix up for the marketing site plugins page. However, the Backstage website is directly linking to some of our assets and now has broken images. 

Screenshot 2020-11-19 at 15 19 50

I'm reverting the change until we can figure out what to do.


dtuite:
It's better to just revert the whole PR instead. https://github.com/RoadieHQ/marketing-site/pull/55

PR #63, opened 2020-11-27
Landing improvements

dtuite:


src/components/home/features/ServiceCatalog.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+}));
+
+const Text = () => {
+  const classes = useStyles();
+
+  return (
+    <div>
+      <h2>Service catalog for discoverability</h2>
+      <p className={classes.p}>300 GitHub repos and no idea what they do?</p>
```
iain-b, line 19: I'm not sure about the line above here. I think it's the "no idea" part. I dunno I reckon customers would like to think they've some idea 😄 Nit pick so feel free to ignore; I don't have a better suggestion.

src/pages/index.js
```diff
@@ -26,7 +47,40 @@ const Home = ({ data, location }) => {
...
 
       <StickyFooter location={location}>
-        <Hero setModalOpen={setModalOpen} siteMetadata={data.site.siteMetadata} />
+        <ResponsiveSpacer>
+          <Hero setModalOpen={setModalOpen} siteMetadata={data.site.siteMetadata} />
+        </ResponsiveSpacer>
+
+        <ResponsiveSpacer>
+          <InterstitialTitle fontSize="3em">
+            Do away with tribal knowledge
```
iain-b, line 56: Just an idea maybe "invert" the suggestion suggest that they can put their tribal knowledge into backstage and _speed up_ development. On the other hand I suppose you're trying to get them to think of the problem we're solving...

iain-b (approved):
I'm going to go out on a limb and say marketing copy isn't my strong suit but I've left a couple of comments. Very minor just things which struck me as I was reading.

dtuite:
Good suggestions @iain-b . I will improve both of these things before merging.

src/components/home/features/ServiceCatalog.js
punkle: I wonder should we mention github if it can be used for other repository servers?

dtuite:
@iain-b I have addressed your comments. I'm going to merge this without a +1 since it's just the marketing site and nothing can go too wrong. It's also the weekend and Christmas and I don't want to wait for a review.

PR #67, opened 2020-12-07
Add argo-cd plugin

mcalus3:


content/plugins/descriptions/argo-cd.yaml
```diff
@@ -0,0 +1,94 @@
...
+# languages used here must be listed in the .babelrc
+gettingStarted:
+  # What will this step accomplish?
+  - intro: Install the plugin into Backstage.
+    language: bash
+    code: yarn add @roadiehq/backstage-plugin-argo-cd
+  - intro: Add proxy config
+    language: yaml
+    code: |
+      // app-config.yaml
```
dtuite, line 44: Valid comments in yaml are `#` rather than double slashes.

content/plugins/descriptions/argo-cd.yaml
```diff
@@ -0,0 +1,94 @@
...
+  - intro: Add annotation to the yaml config file of a component
+    language: yaml
+    code: |
+      metadata:
+        annotations:
+          argocd/app-name: &lt;your-app-name>
+  - intro: Get and provide `ARGOCD_AUTH_TOKEN` as env variable in following format
+    language: yaml
+    code: |
+      `argocd.token=&lt;token>`
```
dtuite, line 88: The backticks are not required here.

It's also slightly confusing that the name of the env var is in the intro, but the token value is in the code sample. Could you instead please provide both the env var name and value in the code snippet.

dtuite (changes requested):

content/plugins/descriptions/argo-cd.yaml
mcalus3: Done

content/plugins/descriptions/argo-cd.yaml
mcalus3: Done

dtuite (approved):

PR #70, opened 2020-12-07
E2e testing

dtuite:
Add cypress.io and two tests which exercise the form on the landing page.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -1,0 +1,25 @@
+// ***********************************************
+// This example commands.js shows you how to
+// create various custom commands and overwrite
+// existing commands.
+//
+// For more comprehensive examples of custom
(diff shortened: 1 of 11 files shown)


cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
martina-if, line 1: Is this file supposed to be commited?

martina-if (dismissed):

cypress/support/commands.js
dtuite: @martina-if I removed it and the build failed. The testing framework seems to expect it to be there, even if it doesn't do anything.

BTW I need a new review now please. 😀

martina-if (approved):

PR #74, opened 2020-12-10
Workaround for images duplication bug

### LINKED ISSUE #52: Incorrect screenshot on GitHub PRs marketing page.
This shows the Security Insights plugin.

https://roadie.io/backstage/plugins/github-pull-requests/

mcalus3:
Closes #52 

There is some kind of a gatsby bug with building images. When you open the site's source, you'll see that the image's src is src="/static/c7a8e87fc6a64f52bf3d2a257210ed60/ee604/GitHub-Mark-120px-plus.png". This is incorrect, somehow the image's source is set after the logo image and that's the real reason why those two github plugins share the same cover image - they have the same icon (in fact, github actions also).

My proposition for solving this is a shortest workaround - I've created three image files with logo to avoid any two plugins using the same file.


mcalus3:
@dtuite Added comments about the workaround, can be merged.

martina-if (approved):

PR #101, opened 2021-02-01
Feature: Kubernetes plugin

ayshiff:
This Draft PR add the Kubernetes plugin entry.
This is mostly inspired by the [Backstage Kubernetes Installation/Configuration guide](https://backstage.io/docs/features/kubernetes/overview).

It also adds a new getting started element: `title`.
I found it useful to make sections in the getting started blocks to better separate the different contexts.
(In this case the frontend and backend plugin).

One problem I encountered was that third level nested unordered list items were rendered as second level elements in markdowns files. I saw that I was not the only one with this problem and that one of the solutions was to to indent the lists with 4 spaces. That's why I added a specific prettier configuration for the kubernetes file.
(I am looking for a better solution to fix this problem.)

I will iterate in the coming days to improve this version.

I'm waiting for your feedback!

**NOTE:** I also take this opportunity to say that it might be interesting to add a github action like the one used by backstage: [vale-ai](https://github.com/errata-ai/vale) to check the quality of the markdown files ([link](https://github.com/backstage/backstage/blob/master/.github/workflows/docs-quality-checker.yml)).    
I think that the tool could be interesting to integrate to ensure a certain quality in our `md` files.


content/plugins/notes/kubernetes.md
```diff
@@ -0,0 +1,207 @@
+---
+name: kubernetes
+---
+
+You have added the two Kubernetes plugins to your Backstage application:
+
+-   **frontend**: it will take care of displaying the information to the user.
+-   **backend**: it will take care of connecting to the Kubernetes clusters and sending the information to the frontend.
+
+After installing the plugings, you have to configure them in two steps:
```
dtuite, line 10: Small typo in `plugings`.

content/plugins/notes/kubernetes.md
```diff
@@ -0,0 +1,207 @@
+---
+name: kubernetes
+---
+
+You have added the two Kubernetes plugins to your Backstage application:
```
dtuite, line 5: Perhaps this should be something like

> The Backstage Kubernetes plugin has two separate components:

content/plugins/notes/kubernetes.md
```diff
@@ -0,0 +1,207 @@
...
+          name: minikube
+          authProvider: 'serviceAccount'
+          serviceAccountToken:
+              $env: K8S_MINIKUBE_TOKEN
+        - url: http://127.0.0.2:9999
+          name: gke-cluster-1
+          authProvider: 'google'
+```
+
+Here is the list of fields in detail:
```
dtuite, line 36: Let's link to [the official docs](https://backstage.io/docs/features/kubernetes/configuration) instead of duplicating this information please. It's too likely to go stale.

content/plugins/notes/kubernetes.md
```diff
@@ -0,0 +1,207 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+    namespace: default
+    name: pod-reader
+rules:
+    - apiGroups: ['']
+      resources: ['pods']
+      verbs: ['get', 'watch', 'list']
+```
```
dtuite, line 116: I love that you added this.

content/plugins/notes/kubernetes.md
```diff
@@ -0,0 +1,207 @@
...
+[Labels and Selectors documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)
+for more info.
+
+```yaml
+# catalog-info.yaml
+annotations:
+    'backstage.io/kubernetes-label-selector': 'app=my-app,component=front-end'
+```
+
+## Example of steps to follow
```
dtuite, line 165: I love that you added an example. Anytime I find myself reading documentation, a simple and clear example that I can follow is what helps me the most.

content/plugins/notes/kubernetes.md
```diff
@@ -0,0 +1,207 @@
...
+rules:
+    - apiGroups: ['']
+      resources: ['pods']
+      verbs: ['get', 'watch', 'list']
+```
+
+## Surfacing your Kubernetes components as part of an entity
+
+There are two ways to surface your Kubernetes components as part of an entity.  
+**NOTE:** The label selector takes precedence over the annotation/service id.
```
dtuite, line 121: Is there a clear reason why I would use one method or the other? If so, we should add that here.

dtuite (approved):
Really great. I spotted one typo.

content/plugins/notes/kubernetes.md
ayshiff: I don't think there's a clear reason why one method should be used over the other. I just know that in the Backstage repository, examples use the annotation/service id.

dtuite (approved):

PR #116, opened 2021-02-23
Use slug for legal notices

dtuite:
There is no customer facing impact from this. It just removes a redundant property from the frontmatter of the legal notices.


dtuite:
https://app.clubhouse.io/larder/story/1312/make-a-place-for-customer-docs

padraigobrien (approved):

PR #117, opened 2021-02-23
Adding in a new job for platform engineer and modifying javascript engineer 

padraigobrien:
- Added in a new career page for platform engineer.
- Modified the javascript engineer page to reflect the pay in euro as opposed to dolars

### DIFF
--- a/src/pages/careers.js
+++ b/src/pages/careers.js
@@ -78,7 +78,14 @@ const OPEN_ROLES = [
       text: `Become a top contributor to the open-source software that Roadie is built on.`,
     },
   ],
-  [],
+  [
+    {
+      title() {
+        return <Link to="/careers/platform-engineer/">Platform Engineer</Link>;
+      },
+      text: `Come build Roadies platform`,
+    },
+  ],
 ];
 
 const Careers = ({ data, location }) => {
--- a/src/pages/careers.js
+++ b/src/pages/careers.js
@@ -83,7 +83,7 @@ const OPEN_ROLES = [
       title() {
         return <Link to="/careers/platform-engineer/">Platform Engineer</Link>;
       },
-      text: `Come build Roadies platform`,
+      text: `Take an active role in shaping the future of our platform`,
     },
   ],
 ];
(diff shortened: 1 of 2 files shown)


src/pages/careers.js
```diff
@@ -78,7 +78,14 @@ const OPEN_ROLES = [
       text: `Become a top contributor to the open-source software that Roadie is built on.`,
     },
   ],
-  [],
+  [
+    {
+      title() {
+        return <Link to="/careers/platform-engineer/">Platform Engineer</Link>;
+      },
+      text: `Come build Roadies platform`,
```
dtuite, line 86: I wonder can we talk about a benefit that the candidate will receive if they take the job here? This particular proposition doesn't jump out at me as extremely attractive.

Grammar wise, should it be "Roadie's platform"? I can never remember this rule.

src/pages/careers/platform-engineer.js
```diff
@@ -0,0 +1,103 @@
...
+  sitewideHeaderWrapper: {
+    marginBottom: 40,
+    paddingLeft: 16,
+    paddingRight: 16,
+  },
+}));
+
+const ROLE_NAME = 'Platform Engineer';
+const TYPEFORM_SLUG = 'platform-engineer';
+const HEADLINE = 'Help companies adopt backstage.io';
```
dtuite, line 22: Personally, this headline makes me think of solutions engineer roles. I would point to the platform being built instead I think.

dtuite (changes requested):

src/pages/careers.js
padraigobrien: how about "take an active role in shaping the future of our platform"

dtuite: Sounds better to me.

padraigobrien: remedied to "Take an active role in shaping the future of our platform"

src/pages/careers/platform-engineer.js
padraigobrien: remedied to "Take an active role in shaping the future of our platform"

dtuite (approved):

PR #120, opened 2021-03-07
Simplify use of gatsby-source-filesystem

dtuite:
I guess I didn't understand how this was supposed to work when I first
started using Gatsby. I had one instance of gatsby-source-filesystem for
each subdirectory of `content`. This would create slugs which didn't
have the correct prefix on them (like `blog/xyz`). I dealt with this by
manually adding the `/blog/` prefix to each URL in `gatsby-node.js`.

A much simpler way is to load all subdirectories of content with one
instance of gatsby-source-filesystem. Then you get the directory names
as part of the URL automatically and the manual prefixing can be removed
from `gatsby-node.js`.

This change is non-user facing. The public URLs of the website should
all remain the same as they are now.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -20,18 +20,10 @@ module.exports = {
     {
       resolve: `gatsby-source-filesystem`,
       options: {
-        path: `${__dirname}/content/blog`,
-        name: `blog`,
+        path: `${__dirname}/content`,
+        name: 'content',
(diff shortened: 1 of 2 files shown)


dtuite:
Merging without review as this is totally internal and I want to rebase it onto some active branches.

I have checked against roadie.io/sitemap.xml and there are no URL changes in this PR.

PR #134, opened 2021-03-13
Version legal notices

### LINKED ISSUE #132: Update sub-processors
Caused by the table.

https://roadie.io/legal-notices/sub-processors/

Also, remove HubSpot which is off-platform.

dtuite:
- Version the legal notices so customers can go back in history and look at changes over time.
- Fix up some styling on the legal notices page. It looks more like the Documentation pages now.
- Fix a problem with tables in the legal notices causing horizontal scrolling on mobile.
- Removes Hubspot as a sub-processor because it is "off platform".

Fixes #132


dtuite:
https://app.clubhouse.io/larder/story/1543/version-legal-policies

src/pageCreation/createLatestLegalNotices.js
```diff
@@ -0,0 +1,54 @@
...
+  const maxVersionForType = get(data, 'notices.edges').reduce((accumulator, edge) => {
+    const matches = edge.node.fields.slug.match(/\/legal-notices\/([\w-]+)\/v(\d+)\//);
+    const noticeType = matches[1];
+    const version = parseInt(matches[2]);
+
+    if (!has(accumulator, noticeType) || version > accumulator[noticeType]) {
+      accumulator[noticeType] = version;
+    } else if (version === accumulator[noticeType]) {
+      throw new Error(
+        `Duplicate version number v${version} found for legal-notice type: ${noticeType}`
```
martina-if, line 36: can this happen given that they are files in a folder?

src/pageCreation/createLatestLegalNotices.js
```diff
@@ -0,0 +1,54 @@
...
+ */
+const createLatestLegalNotices = async ({ graphql, actions: { createPage } }) => {
+  const component = path.resolve(__dirname, '../templates/LegalNotice.js');
+  const { data, errors } = await graphql(LEGAL_NOTICES_QUERY);
+
+  if (errors) {
+    throw errors;
+  }
+
+  const maxVersionForType = get(data, 'notices.edges').reduce((accumulator, edge) => {
```
martina-if, line 27: An alternative implementation that could look simpler to read is to map them first to each type and then sort them?

martina-if (approved):

src/pageCreation/createLatestLegalNotices.js
dtuite: Good point. I can't see a way that it can.

src/pageCreation/createLatestLegalNotices.js
dtuite: That version looks like this. What do you think?

```js
  const versionsForEachType = get(data, 'notices.edges').reduce((accumulator, edge) => {
    const matches = edge.node.fields.slug.match(/\/legal-notices\/([\w-]+)\/v(\d+)\//);
    const noticeType = matches[1];
    const version = parseInt(matches[2]);

    if (has(accumulator, noticeType)) {
      accumulator[noticeType] = [...accumulator[noticeType], version];
    } else {
      accumulator[noticeType] = [version];
    }

    return accumulator;
  }, {});

  const maxVersionForType = reduce(versionsForEachType, (accumulator, versions, noticeType) => {
    accumulator[noticeType] = max(versions);
    return accumulator;
  }, {});
```

martina-if: hm, it doesn't really look much simpler so up to you :shrug: no strong opinion here

dtuite:
@martina-if I removed the unnecessary error checking and rebased.

martina-if (approved):

PR #137, opened 2021-03-15
Add documentation for custom plugins

martina-if:
[Ch1422]

Is this the right place in the sidebar?

plugins_docs

### DIFF
--- a/content/docs/custom-plugins/index.md
+++ b/content/docs/custom-plugins/index.md
@@ -13,7 +13,7 @@ your plugins, and we host a dedicated npm registry which you can publish to.
 
 For our alpha release you'll need to give us some information upfront about your plugin. We plan to make all of this
 self-service going forward but currently Roadie doesn't support this out-of-the-box. For now, we will need you to
-fill out [this form][form] the first time you want to add a new plugin and pushing new versions will not require you
+fill out [this form][form] the first time you want to add a new plugin. Pushing new versions will not require you
 to fill the form again.
 
 ## Pushing the plugin to Roadie
--- a/content/docs/custom-plugins/index.md
+++ b/content/docs/custom-plugins/index.md
@@ -9,21 +9,22 @@ description: How to add a custom Backstage plugin to Roadie
 Your plugins can be published via npm or yarn, like publishing a normal package. We will provide you a unique scope for
 your plugins, and we host a dedicated npm registry which you can publish to.
 
-### Prerequisites
+## Prerequisites
 
 For our alpha release you'll need to give us some information upfront about your plugin. We plan to make all of this
 self-service going forward but currently Roadie doesn't support this out-of-the-box. For now, we will need you to
-fill out [this form][form].
+fill out [this form][form] the first time you want to add a new plugin and pushing new versions will not require you
+to fill the form again.
 
-### Pushing the plugin to Roadie
+## Pushing the plugin to Roadie
 
-#### 1. Get credentials
+### Step 1. Get credentials
 
 Get your credentials for the npm registry. Follow [this link][forgot-password] to reset your password. The username is
 `<your-company>-roadie`. This will send a reset link to the email we have associated with your account. You can
 change this email as you wish.
 
-#### 2. Configure NPM
+### Step 2. Configure NPM
 
 To configure NPM there are two options:
 
@@ -42,7 +43,7 @@ To configure NPM there are two options:
   always-auth = true
   </pre>
 
-#### 3. Publishing
+### Step 3. Publishing
 
 Once you're ready to publish your plugin these are the steps you'll need to follow:
 
@@ -52,6 +53,12 @@ Once you're ready to publish your plugin these are the steps you'll need to foll
 3. Build your plugin (e.g. yarn install && yarn tsc && yarn build)
 4. Publish your plugin npm publish
 5. (Optional) Check the registry npm info `@<your-company>-roadie/<package-name>`
+
+## Conclusion
+
+We will notify you via Slack once the plugin is ready in your Roadie Backstage instance. Further updates to the npm
+package will be picked up and released automatically and you can expect the changes to appear in Backstage after 15-20
+minutes.
 
 [forgot-password]: https://roadiehq.jfrog.io/ui/login/forgot-password
 [form]: https://docs.google.com/forms/d/e/1FAIpQLSdSNr4Ps_RpKEx0V2QbxWaKLb3-DKi0W7U09Wth0SXHQoPyXQ/viewform
--- a/content/docs/custom-plugins/index.md
+++ b/content/docs/custom-plugins/index.md
@@ -1,0 +1,51 @@
+---
+title: Installing Custom Plugins
+lastUpdated: '2021-03-16T17:55:00.0Z'
+description: How to add a custom Backstage plugin to Roadie
+---
+
+# Introduction
+
+Your plugins can be installed via npm or yarn, like publishing a normal package. We will provide you a unique scope for
+your plugins, and we host a dedicated npm registry which you can publish to.
+
+## Prerequisites
+
+- For our alpha release you'll need to give us some information upfront about your plugin. We plan to make all of this
+  self-service going forward but currently backstage doesn't support this out-of-the-box. For now, we will need you to
+  fill out [this form][form].
+
+- Get your credentials for the npm registry. Follow [this link][forgot-password] to reset your password. The username is
+  `<your-company>-roadie`. This will send a reset link to the email we have associated with your account. You can
+  change this email as you wish.
+
+- Configure npm. There are two options:
+
+  - Interactively using npm login. Suitable for a developer testing the process.
+    <pre>
+    npm config set @<b>&lt;your-company&gt;</b>-roadie:registry https://roadiehq.jfrog.io/artifactory/api/npm/<b>&lt;your-company&gt;</b>-roadie/
+    npm login --scope=@<b>&lt;your-company&gt;</b>-roadie 
+    # This will prompt for your username, password and email
+    </pre>
+
+  - In a CI/CD environment you can write a local `.npmrc` file
+    <pre>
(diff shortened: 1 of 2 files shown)


dtuite:
@martina-if You know the build failed right? All good if so. 👍

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Installing Custom Plugins
+lastUpdated: ''
```
dtuite, line 3: @martina-if The build will pass if you put an ISO formatted timestamp in here. e.g. `2021-03-16T21:00:00.0Z`

The time portion will be discarded but the date portion is used to display the doc was "published" on. Like this:

Screenshot 2021-03-16 at 16 21 07

Ideally it would say "Last updated at [date]" instead of "published" on but it's hopefully good enough for now.

martina-if: Thanks! fixed

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Installing Custom Plugins
+lastUpdated: '2021-03-16T17:55:00.0Z'
+description: How to add a custom Backstage plugin to Roadie
+---
+
+# Introduction
+
+Your plugins can be installed via npm or yarn, like publishing a normal package. We will provide you a unique scope for
```
dtuite, line 9:
```suggestion
Your plugins can be published via npm or yarn, like publishing a normal package. We will provide you a unique scope for
```

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Installing Custom Plugins
+lastUpdated: '2021-03-16T17:55:00.0Z'
+description: How to add a custom Backstage plugin to Roadie
+---
+
+# Introduction
```
dtuite, line 7: This should be a h2 to match the other docs we have.

```suggestion
## Introduction
```

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
...
+
+Your plugins can be installed via npm or yarn, like publishing a normal package. We will provide you a unique scope for
+your plugins, and we host a dedicated npm registry which you can publish to.
+
+## Prerequisites
+
+- For our alpha release you'll need to give us some information upfront about your plugin. We plan to make all of this
+  self-service going forward but currently backstage doesn't support this out-of-the-box. For now, we will need you to
```
dtuite, line 15: I would blame Roadie for this rather than Backstage. It's unlikely Backstage will ever support this workflow.

```suggestion
self-service going forward but currently Roadie doesn't support this out-of-the-box. For now, we will need you to
```

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
...
+  fill out [this form][form].
```
dtuite, line 16: 🤯  I didn't know you could do this in markdown. This is awesome!

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
...
+
+- Get your credentials for the npm registry. Follow [this link][forgot-password] to reset your password. The username is
```
dtuite, line 18: Can we split out separate steps please, like the headings in the existing [Sentry docs](https://deploy-preview-137--roadie.netlify.app/docs/integrations/sentry/).

It looks like we have steps for

1. Get credentials
2. Configure NPM
3. Publish

Separate headings have the 2 advantages:

1. They are linkable so it's easier to send someone to a specific spot in the docs
2. We can automatically extract a table of contents in the future and make a sidebar on the right side of the page for easier navigation.

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
...
+
+Once you're ready to publish your plugin these are the steps you'll need to follow:
+
+1. Ensure your package is scoped correctly in package.json. The name field should follow this pattern
+   `@<your-company>-roadie/<package-name>`
+2. Select a new package version to publish. You can't overwrite a version once published (e.g. npm version patch)
+3. Build your plugin (e.g. yarn install && yarn tsc && yarn build)
+4. Publish your plugin npm publish
+5. (Optional) Check the registry npm info `@<your-company>-roadie/<package-name>`
+
```
dtuite, line 49: Is there a way to verify that everything worked or something we can tell the user to do as a next step? It feels kinda inconclusive down here.

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,51 @@
...
+  fill out [this form][form].
+
+- Get your credentials for the npm registry. Follow [this link][forgot-password] to reset your password. The username is
+  `<your-company>-roadie`. This will send a reset link to the email we have associated with your account. You can
+  change this email as you wish.
+
+- Configure npm. There are two options:
+
+  - Interactively using npm login. Suitable for a developer testing the process.
+    <pre>
```
dtuite, line 25: You'll get code formatting and syntax highlighting here if you use backticks with the name of the language.

dtuite (changes requested):
You'll need to manually edit the sidebar of the documentation to add a link to your page. Let's make a new section called "Advanced" in the sidebar and put this page in there.

https://github.com/RoadieHQ/marketing-site/blob/main/src/components/doc/Sidebar.js

The directory structure for this page of docs should then be `content/docs/advanced/custom-plugins/index.md`.

Also, this needs a rebase.

content/docs/custom-plugins/index.md
martina-if: :+1: good point. Done

content/docs/custom-plugins/index.md
martina-if: AFAIK we don't have anything for that yet. I added this story to the custom plugin epic https://app.clubhouse.io/larder/story/1571/give-user-feedback-on-publishing-operation

This is for the old way of adding a custom plugin so we should decide if we want to build that or just add the feedback in the new way of adding custom plugins. I will ask Iain anyways in case I missed something.

content/docs/custom-plugins/index.md
martina-if: I used `<pre>` because I want to highlight in bold the test that needs to be replaced. Should I change it back and remove the highlighting?

pre_bold_highlight

with backticks:
backticks

content/docs/custom-plugins/index.md
```diff
@@ -0,0 +1,57 @@
...
+lastUpdated: '2021-03-17T12:11:00.0Z'
+description: How to add a custom Backstage plugin to Roadie
+---
+
+## Introduction
+
+Your plugins can be published via npm or yarn, like publishing a normal package. We will provide you a unique scope for
+your plugins, and we host a dedicated npm registry which you can publish to.
+
+### Prerequisites
```
dtuite, line 12: My suggested heading structure is this:

```plain
## Introduction

## Prerequisites

## Pushing the plugins to Roadie

### Step 1: Get credentials

### Step 2: Configure NPM

### Step 3: Publishing

## Conclusion
```

The automatically generated table of contents will include any heading which is h2. and ignore h3., h4. etc. So the table of contents will be:

```plain
1. Introduction
2. Prerequisites
3. Pushing the plugins to Roadie
4. Conclusion
```

My inspiration for this is the [Hasura docs](https://hasura.io/docs/latest/graphql/core/remote-schemas/adding-schema.html). The main difference being that we don't have so many granular pages yet (or a good way to navigate between them) so it makes sense to consolidate documentation pages more than Hasura have done.

content/docs/custom-plugins/index.md
dtuite: Perhaps we could explain how we will notify them when their custom plugin shows up?

It might also be worth saying that future updates to the plugin won't require re-submission of the form or manual steps and thus should be must faster.

dtuite (changes requested):

content/docs/custom-plugins/index.md
martina-if: Not sure what you mean by this
> Perhaps we could explain how we will notify them when their custom plugin shows up?

We can't see when it shows up. Should we say, we will notify them on slack when the first build is ready? 

I will add now the second part you mention

content/docs/custom-plugins/index.md
dtuite: I guess that makes sense. I have opened https://github.com/RoadieHQ/marketing-site/issues/142 for more work in this area.

content/docs/custom-plugins/index.md
martina-if: This is a very short document with a few steps :thinking: What should I say in the conclusion section?

content/docs/custom-plugins/index.md
dtuite: Yeah exactly. We haven't told them what happens after they submit the Google form, so they won't know when the plugin is installed and ready to use. We should tell them that we will notify them via their shared Slack channel or via email when the plugin is ready to use after being installed the first time.

content/docs/custom-plugins/index.md
dtuite: I would just summarize what has happened. Something like

> Once the steps above have been completed successfully and you have been notified that your plugin is ready to use, you should be able to log into your Backstage instance and use the plugin just like any other plugin. 

> You can update your plugin at will and the changes will appear in Backstage. The form is only required when installing a new plugin for the first time.

martina-if:
I added the conclusion section but it feels like the wrong word for this. Perhaps "next steps" or "what to expect next" or something similar would be better.

content/docs/custom-plugins/index.md
martina-if: Added some comments to the prerequisites, and the end of the file. Let me know what you think about those

dtuite (approved):

dtuite:
The `pre` tag is causing horizontal scrolling on mobile but most of our traffic is desktop so we can fix it later.

dtuite (approved):

PR #152, opened 2021-03-23
Split cookies policies

dtuite:
I have been advised by counsel to split the cookies into two categories: those that exist in the application and those that exist on the website.


dtuite:
https://app.clubhouse.io/larder/story/1641/split-cookie-policy

Xantier (approved):

PR #166, opened 2021-04-01
add docs for creating a pager duty token

punkle:
[ch1536]


content/docs/integrations/pagerduty/index.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Generating PagerDuty Token
+lastUpdated: '2021-03-16T21:00:00.0Z'
```
dtuite, line 3: It's fine for now but ideally the `lastUpdated` date should be changed to today. Recent dates give people confidence that the docs are actually correct and kept updated.

dtuite (approved):

PR #174, opened 2021-04-06
Update index.md

iain-b:
These docs are incorrect we require the name to be roadie-backstage-admin


iain-b:
I'll re-do the images too so they have the right group name. I wonder if we should start by saying you need to create an admin group called "roadie-backstage-admin" and then give them manual instructions but sort of leave it open to other ways of importing groups?

dtuite (dismissed):

PR #192, opened 2021-04-22
Skip Algolia indexing in preview deploys

dtuite:
This should hopefully speed up Netlify preview builds by skipping the indexing of documents in Algolia.

The index will be updated when the producton deploy runs on Netlify.

This is the behavior we want anyway. Currently, draft documents are appearing in production search because they are indexed when Netlify's preview build runs.


dtuite:
https://app.clubhouse.io/larder/story/2113/speed-up-marketing-site-builds

punkle (approved):

PR #209, opened 2021-05-10
Force a trailing slash on internal links

dtuite:
This is a continuation of https://github.com/RoadieHQ/marketing-site/pull/94, or of `9923f1df2957370afe70a303d101af289e46d4bd` more specifically.

The plugin added in that PR was only part of the requirements. We also need to do step 3 described on [this
page](https://www.gatsbyjs.com/plugins/gatsby-plugin-force-trailing-slashes/?=Link).

> All uses of Gatsby’s <Link> component in your code to intentionally specify a trailing slash.

This commit does that.

[This guide](https://attacomsian.com/blog/javascript-url-object) was useful for learning that I can mutate `URL` objects.


src/components/Link/Link.js
```diff
@@ -11,9 +12,14 @@ const Link = ({
   ...rest
 }) => {
   if (to.startsWith('/') || to.startsWith('#')) {
+    let internalUrl = new URL(to, siteMetadata.siteUrl);
```
dtuite, line 15: The second argument passed into `URL` here is:

1. Incorrect in environments other than production - eg. local development or Netlify preview environments.
2. Breaking hash URLs, such as those in [the right sidebar of docs](https://deploy-preview-209--roadie.netlify.app/docs/getting-started/getting-started-for-admins/) pages.

This needs more thought.

Xantier (approved):
SEO seems either flaky or black magic every time

PR #222, opened 2021-05-20
Remove old careers site and redirect to the new one

martina-if:


src/components/SitewideHeader/HamburgerMenu.js
```diff
@@ -126,7 +126,7 @@ const HamburgerMenu = ({ siteMetadata }) => {
 
         <div className={classes.spacer}>
           <TextLink
-            to="/careers/"
+              to="/careers/"
```
dtuite, line 129: Looks to be an errant indentation here.

Why not link directly to the new site and save the reader a network hop?

```
<TextLink to="https://careers.roadie.io" ...
```

dtuite (commented):
Needs a rebase

src/components/SitewideHeader/HamburgerMenu.js
martina-if: Ah, because I thought you suggested to add a 302 in the Netlify config instead. I can do both :+1:

dtuite: You need both. 

The change to this link above is for users who are already on the website and clicking around between different pages. When they click the footer link we need to send them over to the careers page. 

The 301 redirect in the netlify config is to redirect users who are incoming directly to the old careers path (perhaps from Google or somewhere else).

martina-if:
I had only removed the pages but not the components, so now I removed everything in `src/components/careers`.

dtuite (approved):

PR #225, opened 2021-05-21
update github pull-requests and github-plugin

Irma12:
Updated instructions for :
Firebase-functions
Github insights
Github Pull Requests

### DIFF
--- a/content/backstage/plugins/firebase-functions.md
+++ b/content/backstage/plugins/firebase-functions.md
@@ -26,17 +26,52 @@ gettingStarted:
     language: 'ts'
     code: |
       // packages/app/src/plugins.ts
-      export { plugin as FirebaseFunctionsPlugin } from '@roadiehq/backstage-plugin-travis-ci';
+      export { plugin as FirebaseFunctionsPlugin } from '@roadiehq/backstage-plugin-firebase-functions';
 
-  - intro: 'Add plugin to the `entitytPage.tsx` source file:'
+  - intro: 'Add plugin to the `entityPage.tsx` source file:'
     language: typescript
     code: |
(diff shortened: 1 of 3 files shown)


content/backstage/plugins/firebase-functions.md
```diff
@@ -26,17 +26,52 @@ gettingStarted:
     language: 'ts'
     code: |
       // packages/app/src/plugins.ts
```
dtuite, line 28: Has this plugin actually been updated to use the composability API? I ask, because this step of adding the plugin to the `src/plugins.ts` page seems to not be needed in the other plugins I've looked at which have been migrated.

content/backstage/plugins/github-insights.md
```diff
@@ -25,25 +25,24 @@ gettingStarted: # What will this step accomplish?
     language: typescript
     code: |
       // packages/app/src/plugins.ts
-      export { plugin as GitHubInsights } from '@roadiehq/backstage-plugin-github-insights';
```
dtuite, line 28: I'm 90% sure that this plugin has been upgraded and this step shouldn't be needed any more. Perhaps the plugin README just wasn't updated?

Irma12: @dtuite You are right, I will also correct it in README

dtuite: Cool. Other than that this is looking pretty good.

dtuite (approved):
Looks good. Thanks

We should do another round of reviews for these pages some time soon. The pull requests plugin, for example, should talk about authentication and OAuth as part of the guide.

PR #235, opened 2021-05-27
Update buildkite plugin configuration

nicolasarnold12321:
[ch2425]

### DIFF
--- a/content/backstage/plugins/buildkite.md
+++ b/content/backstage/plugins/buildkite.md
@@ -23,12 +23,6 @@ gettingStarted:
     language: bash
     code: yarn add @roadiehq/backstage-plugin-buildkite
 
-  - intro: Import it into your Backstage application
-    language: typescript
-    code: |
-      // packages/app/src/plugins.ts
-      export { plugin as Buildkite } from '@roadiehq/backstage-plugin-buildkite';
-
   - intro: Add proxy configuration'
     language: YAML
     code: |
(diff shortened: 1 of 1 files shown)


content/backstage/plugins/buildkite.md
```diff
@@ -40,30 +34,39 @@ gettingStarted:
             Authorization: 
               $env: BUILDKITE_TOKEN
 
+  - intro: Add plugin to the list of your plugins
+    language: typescript
```
iain-b, line 38: This step is no longer necessary AFAIK

content/backstage/plugins/buildkite.md
```diff
@@ -40,30 +34,39 @@ gettingStarted:
...
 
-      const CICDSwitcher = ({ entity }: { entity: Entity }) => {
-        switch (true) {
-          ...
-          case isBuildkiteAvailable(entity):
-            return <BuildkiteRouter entity={entity} />;
-          ...
-        }
-      }
+     const CICDSwitcher = ({ entity }: { entity: Entity }) => {
```
iain-b, line 52: This is a bit out of date. While it'll work we should push users to use the new catalog API. You can see an example in the upstream backstage (packages/app/src/components/catalog/EntityPage.tsx):
```
export const cicdContent = (
  <EntitySwitch>
    <EntitySwitch.Case if={isJenkinsAvailable}>
      <EntityJenkinsContent />
    </EntitySwitch.Case>

    <EntitySwitch.Case if={isBuildkiteAvailable}>
      <EntityBuildkiteContent />
    </EntitySwitch.Case>
    ```

content/backstage/plugins/buildkite.md
nicolasarnold12321: Ah ok cool! I had just taken it from here: https://github.com/RoadieHQ/backstage-plugin-buildkite/blob/main/README.md. I'll remove it now so.

iain-b (approved):

PR #241, opened 2021-05-29
Add Netlify CMS

dtuite:
Just an investigation for the moment.

Potential benefits:

1. Easier publishing workflow.
2. Non-technical users can publish
3. Can show "last updated" dates on various pages without having to manually track.


dtuite:
Closing this. My experience is that the Netlify CMS doesn't have any special tricks that another CMS would have, and it is likely not as well supported as a dedicated CMS.

PR #246, opened 2021-06-06
Add evaluation licence

dtuite:
These are the short terms that prospects will accept when they evaluate Roadie.

[Direct link to preview](https://deploy-preview-246--roadie.netlify.app/legal-notices/evaluation-licence/v1/)

[ch2600]


oblongcoffee (approved):
Fixed incorrect clause reference and added 12 month retention period.

PR #266, opened 2021-06-18
Fix search

dtuite:
Search is not working in production. This fixes it and adds an integration test for searching.


martina-if (commented):
LGTM

PR #269, opened 2021-06-21
Update-github-pr-readme

Irma12:


content/backstage/plugins/github-pull-requests.md
```diff
@@ -62,3 +64,15 @@ gettingStarted:
           </Grid>
         );
 ---
+
+
+## Repository migration notice
+
+In order to make testing and deployment of our plugins easier we are migrating all Roadie plugins to a monorepo at https://github.com/RoadieHQ/backstage-roadie-plugins.
+The plugins will still be published to the same place on NPM and will have the same package names so nothing should change for consumers of these plugins.
```
dtuite, line 72: Newline needed between this paragraph and the one above it.

```suggestion

The plugins will still be published to the same place on NPM and will have the same package names so nothing should change for consumers of these plugins.
```

content/backstage/plugins/github-pull-requests.md
```diff
@@ -62,3 +64,15 @@ gettingStarted:
           </Grid>
         );
 ---
+
+
+## Repository migration notice
```
dtuite, line 69: Without a date associated with this text, it's impossible to know if this migration is happening today, if it's scheduled for a month from now, or if it actually already happened and we forgot to take down the notice.

```suggestion
## Repository migration notice (June/July 2021)
```

Also we need a reminder system to make sure we don't forget to take these notices down at the appropriate time.

Irma12: @dtuite Agree, we can keep it non-merged until we publish the new version and make sure it works

dtuite (approved):
@Irma12 I'm approving this. If you want to hold it back, I think you can Request Changes on your own PR to prevent it being accidentally merged?

PR #270, opened 2021-06-21
Add a security.txt file

dtuite:
[ch2832]

### DIFF
--- a/static/security.txt
+++ b/static/security.txt
@@ -1,0 +1,4 @@
+Contact: mailto:security@roadie.io
+Expires: 2022-06-21T11:00:00.000Z
+Preferred-Languages: en,es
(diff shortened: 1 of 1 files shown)


punkle (approved):

static/security.txt
```diff
@@ -0,0 +1,4 @@
+Contact: mailto:security@roadie.io
+Expires: 2022-06-21T11:00:00.000Z
+Preferred-Languages: en,es
```
martina-if, line 3: :smile:

martina-if (approved):

static/security.txt
dtuite: 😉  I should have added Finnish and Bosnian but it's too late now.

PR #274, opened 2021-06-21
moves security.txt into .well-known directory

punkle:
I think it should be in the .well-known directory. See the following link.

https://securitytxt.org/


dtuite (approved):
Oops

PR #299, opened 2021-07-05
Get instance is main CTA

dtuite:
[ch3018]

### DIFF
--- a/src/pages/get-instance.js
+++ b/src/pages/get-instance.js
@@ -3,10 +3,10 @@ import { graphql } from 'gatsby';
 import { SEO } from 'components';
 import { Helmet } from 'react-helmet';
 
-const TYPEFORM_ID = 'tydAeJns';
+const TYPEFORM_ID = 'tR5Cpfis';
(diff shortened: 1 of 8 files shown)


src/pages/get-instance.js
```diff
@@ -0,0 +1,50 @@
...
+const GetInstanceRequest = ({ data, location }) => {
+  const siteTitle = data.site.siteMetadata.title;
+
+  const params = new URLSearchParams(location.search.substring(1));
+  const referredEmail = params.get('referred_email');
+  const clickedButtonLabel = params.get('clicked_button_label');
+  const hashParams = `referred_email=${referredEmail}&clicked_button_label=${clickedButtonLabel}`;
+
+  return (
+    <>
```
punkle, line 18: Whats this do?

dtuite: Which bit? The `<>`?

It groups react elements without adding an actual HTML tag. It's called a fragment. https://reactjs.org/docs/fragments.html

punkle (approved):

PR #304, opened 2021-07-10
Add plausible redirect

dtuite:
[ch3131]


dtuite:
Merging this without a +1 because it's blocking some changes I have queued up in Google Tag Manager.

PR #339, opened 2021-08-05
Make dedicated developer portal landing page

dtuite:
Orla and I are doing a Google Ads conversion rate experiment. We want to see how high we can get conversion rates with a dedicated landing page which targets one search query group.

This PR adds a page which is a clone of the normal landing page but with copy dedicated to the "developer portal" query.

The page is available at `/developer-portal/` and is not linked from anywhere else on the site. So only people with the direct path can find it (not that it is secret or anything).

Preview link: https://deploy-preview-339--roadie.netlify.app/developer-portal/

### DIFF
--- a/src/pages/developer-portal.js
+++ b/src/pages/developer-portal.js
@@ -1,0 +1,86 @@
+import React from 'react';
+import { graphql } from 'gatsby';
+import {
+  SEO,
+  InterstitialTitle,
+  StickyFooter,
+  ResponsiveSpacer,
+} from 'components';
+// eslint-disable-next-line import/named
+import { Testimonial, Hero, FooterCTA, features } from 'components/home';
+
+const SEO_TITLE = 'Backstage Developer Portal';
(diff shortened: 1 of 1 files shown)


src/pages/developer-portal.js
```diff
@@ -0,0 +1,86 @@
...
+import {
+  SEO,
+  InterstitialTitle,
+  StickyFooter,
+  ResponsiveSpacer,
+} from 'components';
+// eslint-disable-next-line import/named
+import { Testimonial, Hero, FooterCTA, features } from 'components/home';
+
+const SEO_TITLE = 'Backstage Developer Portal';
```
dtuite, line 12: @oblongcoffee all the titles and main copy of the page are gathered here in these strings. If you want to make other pages, simply copy/paste this one within the `src/pages` directory and change these strings.

oblongcoffee (approved):
The text is nice and punchy 🚀

PR #353, opened 2021-08-24
Update policy example to include external id

punkle:
[ch3220]

### DIFF
--- a/content/docs/integrations/kubernetes/index.md
+++ b/content/docs/integrations/kubernetes/index.md
@@ -98,7 +98,7 @@ You should see a page like this
       "Action": "sts:AssumeRole",
       "Condition": {
         "StringEquals": {
-          "sts:ExternalId": "XXXXYour ID XXXX"
+          "sts:ExternalId": "EXTERNAL ID AS CONFIGURED ABOVE"
         },
         "StringLike": {
           "aws:PrincipalArn": [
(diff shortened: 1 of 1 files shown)


dtuite:
@punkle I've retriggered the Netlify preview deploy for you. There seemed to be a timeout in Netlify.

content/docs/integrations/kubernetes/index.md
```diff
@@ -97,6 +97,9 @@ You should see a page like this
       },
       "Action": "sts:AssumeRole",
       "Condition": {
+        "StringEquals": {
+          "sts:ExternalId": "XXXXYour ID XXXX"
```
dtuite, line 101: Does AWS generate the External ID for the user or where does it come from?

We use text like `ACCOUNT ARN SUPPLIED ON THE CONFIGURATION PAGE` above and below this line to indicate the place to replace the ARN and the role from AWS. Would it make sense to use the same style of text here instead of `XXXXYour ID XXXX`.

dtuite:
@punkle Netlify appears to be having trouble at the moment. It's no big deal, we can merge this without seeing the preview on Netlify. But just so you know.

content/docs/integrations/kubernetes/index.md
punkle: The External ID is user defined.

punkle: Done, take a look now, let me know what you think.

dtuite (approved):

PR #354, opened 2021-08-25
Add Tech Radar config docs

Xantier:
Story details: https://app.clubhouse.io/larder/story/3662


src/components/doc/Sidebar.js
```diff
@@ -139,6 +139,7 @@ const DocSidebar = () => {
           <SidebarItem to="/docs/integrations/kubernetes/" text="Kubernetes" />
           <SidebarItem to="/docs/integrations/github-token/" text="GitHub via Token" />
           <SidebarItem to="/docs/integrations/sonarqube/" text="SonarQube" />
+          <SidebarItem to="/docs/integrations/tech-radar/" text="Tech Radar" />
```
iain-b, line 142: It's grand for now but this doesn't feel quite right in the integrations section. (though much better to have it than not and no need to create a new section just for this. I'm also not sure what the section would be but perhaps "Configuration" ?)

PR #371, opened 2021-09-05
Send build branch name to analytics services

dtuite:
[ch3890]

### DIFF
--- a/gatsby-ssr.js
+++ b/gatsby-ssr.js
@@ -1,25 +1,10 @@
 const React = require('react');
 const fs = require('fs');
-const map = require('lodash/map');
-
-const { FORM_NAMES } = require('./src/contactFormConstants');
 
 const HEAD_CSS = fs.readFileSync('src/stylesheets/head-loaded-styles.css').toString();
 
-// <form netlify... /> is here to trigger the netlify bot into making form submissions available
-// for this website. If I put the tags on the form which actually performs the submission
-// then it doesn't work. I believe that form is rendered too dynamically to trigger the netlify
-// bot. This form doesn't do any actual work on the site for users.
-const embeddedForms = map(FORM_NAMES, (value) => (
-  <form name={value} netlify="true" netlify-honeypot="bot-field" hidden key="form">
-    <input type="email" name="email" />
-  </form>
-));
-
-exports.onRenderBody = ({ setHeadComponents, setPostBodyComponents }) => {
+exports.onRenderBody = ({ setHeadComponents }) => {
   // Getting these into the head so the fonts start loading fast and we don't have a FoUC.
   // Have to use dangerouslySetInnerHTML to prevent escaping of quotation marks in the CSS.
   setHeadComponents([<style key="inline-styles" dangerouslySetInnerHTML={{ __html: HEAD_CSS }} />]);
-
-  setPostBodyComponents(embeddedForms);
 };
--- a/src/components/actions/EmailCaptureForm.js
+++ b/src/components/actions/EmailCaptureForm.js
@@ -3,6 +3,7 @@ import { createUseStyles } from 'react-jss';
 import classnames from 'classnames';
 import { FaPaperPlane } from 'react-icons/fa';
 
+import { currentlyExecutingGitBranch } from '../../environment';
 import Button from '../Button';
 
 const styles = (theme) => ({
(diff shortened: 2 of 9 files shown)


gatsby-ssr.js
```diff
@@ -1,25 +1,10 @@
...
-// <form netlify... /> is here to trigger the netlify bot into making form submissions available
-// for this website. If I put the tags on the form which actually performs the submission
-// then it doesn't work. I believe that form is rendered too dynamically to trigger the netlify
-// bot. This form doesn't do any actual work on the site for users.
-const embeddedForms = map(FORM_NAMES, (value) => (
-  <form name={value} netlify="true" netlify-honeypot="bot-field" hidden key="form">
-    <input type="email" name="email" />
-  </form>
-));
-
```
dtuite, line 18: I recently learned that none of this stuff is required when using Gatsby with Netlify. Removing it makes this PR simpler because I'm sending params other than `email` to Netlify forms. If I don't remove all this stuff, there is an expectation that these other params would also be added to this static form, in addition to the React form in `NetlifyFormCallToAction.js`

src/googleAnalytics.js
```diff
@@ -0,0 +1,14 @@
+import isUndefined from 'lodash/isUndefined';
+import { trackCustomEvent } from 'gatsby-plugin-google-analytics';
+
+const trackGoogleAnalyticsEvent = (eventParams) => {
+  if (!isUndefined(window.ga) && `"${process.env.BRANCH}"`.length > 0 ) {
+    window.ga('send', 'pageview', {
+      'dimension1': `'${process.env.BRANCH}'`,
```
dtuite, line 7: This dimension data property has been manually configured in GA already.

src/plausible.js
```diff
@@ -1,8 +1,17 @@
...
+import set from 'lodash/set';
 
+const BUILD_TIME_BRANCH_NAME = process.env.BRANCH;
+
+// The plausible tracking snippet is injected into the page using Netlify snippet
+// injection. https://app.netlify.com/sites/roadie/settings/deploys#snippet-injection
+// It serves plausible from a custom domain, which redirects to plausible.io. This
+// redirect is configured in the netlify.toml file.
 const trackPlausibleEvent = (eventName, opts = {}) => {
+  const options = set(opts, 'path.branch', BUILD_TIME_BRANCH_NAME);
```
dtuite, line 11: Will only set at this object path if it doesn't already exist.

src/components/actions/EmailCaptureForm.js
```diff
@@ -95,7 +97,17 @@ const EmailCaptureForm = ({
 
   /* eslint-disable jsx-a11y/no-autofocus */
   return (
-    <form onSubmit={onSubmit}>
+    <form
+      onSubmit={onSubmit}
+      name={netlifyFormName}
+      method="post"
+      data-netlify="true"
+      data-netlify-honeypot="bot-field"
```
dtuite, line 105: Need to add these attributes since we have removed code from `gatsby-ssr.js`.

src/components/actions/NetlifyFormCallToAction.js
```diff
@@ -37,7 +51,7 @@ const NetlifyFormCallToAction = ({
   placeholderText = 'Work email',
   buttonText = 'Click here',
   subFormMessage = 'We will never sell or share your email address.',
-  netlifyFormName = FORM_NAMES.notifyMe,
+  netlifyFormName = FORM_NAMES.getInstance,
```
dtuite, line 54: The `notifyMe` form is legacy so I am updating this default to a form we more recently used.

src/googleAnalytics.js
```diff
@@ -0,0 +1,16 @@
+import isUndefined from 'lodash/isUndefined';
+import { trackCustomEvent } from 'gatsby-plugin-google-analytics';
+
+import { currentlyExecutingGitBranch } from './environment';
+
+const trackGoogleAnalyticsEvent = (eventParams) => {
+  if (!isUndefined(window.ga)) {
```
Xantier, line 7: you could check the existence of the acceptance cookie in here as well I think

Xantier (approved):

src/googleAnalytics.js
dtuite: That would only prevent custom events from being tracked. There must be a better way to turn the whole `ga` object into a no-op. I'll open a separate issue.

PR #394, opened 2021-09-23
Add bugsnag on roadie marketplace

Irma12:
Preview link: https://deploy-preview-394--roadie.netlify.app/backstage/plugins/bugsnag/


content/backstage/plugins/bugsnag.md
```diff
@@ -0,0 +1,86 @@
...
+
+  - intro: Run the backstage app with the following command and navigate to the services tab.
+    code: |
+      yarn start
+
+---
+
+## How to use Bugsnag plugin in Backstage:
+
+1.  Add an annotation to the yaml config file of a component:
```
dtuite, line 65: I'd suggest dropping the numbered list here. It's a bit weird how the line with the number on it is indented and the rest are not. There's only two steps so simply saying "Firstly, add an annotation..." will be ok.

dtuite (dismissed):
Looks good. Just one suggestion about the numbered list in the notes.

dtuite (approved):

PR #395, opened 2021-09-23
Add list of plugin logos to landing page

dtuite:
https://app.shortcut.com/larder/story/4195/display-list-of-plugin-logos

Preview: https://deploy-preview-395--roadie.netlify.app/#integrations

### DIFF
--- a/src/components/home/IntegrationsList.js
+++ b/src/components/home/IntegrationsList.js
@@ -1,0 +1,60 @@
+import React from 'react';
+import { createUseStyles } from 'react-jss';
+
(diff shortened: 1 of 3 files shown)


Xantier:
It looks like there should be more there but I guess the list isn't that long yet

Xantier (approved):

Irma12:
@dtuite I just merged Bugsnag, should we include that one also?

dtuite:
@Irma12 Sure, but it can be added as a separate commit.

PR #404, opened 2021-10-05
Add Prometheus to plugins page.

Xantier:
Preview: https://deploy-preview-404--roadie.netlify.app/backstage/plugins/prometheus/


punkle (approved):
Should we also add a page to the roadie docs?

Xantier:
> Should we also add a page to the roadie docs?

Yup, I'll add the needed config options to Roadie first

PR #455, opened 2021-10-28
Add customer logos to website

dtuite:
https://app.shortcut.com/larder/story/4694/add-customer-logos-to-website


dtuite:
> Just noticed the webp ending on these. I guess there's nowadays no need to support alternative formats for garbage browsers from garbage companies (like Safari for example)?

[caniuse.com](https://caniuse.com/?search=webp) says webp is supported in Safari as long as it's running macOS Big Sur or later. This was released in Nov 2020. 

Unfortunately, it seems that Big Sur is incorrectly reporting itself as Catalina (which is the name of the previous version), so I can't find a stats website which can accurately tell me the % of Big Sur out there. 

54% of our total traffic is macOS (Descktop). If 80% of that was Big Sur, that would leave 11% of our traffic unable to render webp images. That actually does seem to be a problem.

I will fall back to png.

PR #459, opened 2021-11-01
Add travis ci documentation

nicolasarnold12321:


nicolasarnold12321:
https://deploy-preview-459--roadie.netlify.app/docs/integrations/travis-ci/

content/docs/integrations/travis-ci/index.md
```diff
@@ -0,0 +1,39 @@
...
+### Step 2: Store the token in Roadie
+Visit
+```text
+https://<tenant-name>.roadie.so/administration/settings/secrets
+``` and enter the token value from above into `TRAVISCI_AUTH_TOKEN`.
+
+
+## References
+
+- [Token, Token, Token](https://blog.travis-ci.com/2013-01-28-token-token-token)
```
martina-if, line 39: This part is not rendering properly. I think you need a new line after line 31 and on line 43 after the <pre>```</pre>


travis

nicolasarnold12321:
@dtuite I have never come across the CISwitcher. Could you link me some documentation to it? I'll add resources for the annotations so.

dtuite:
@nicolasarnold12321 I might be calling it the wrong thing.

If I read these docs, I would understand how to configure the plugin, but I wouldn't know where to go to actually see it in Roadie. I think we should tell the reader how to accomplish that. That's the next thing they're going to want to do.

nicolasarnold12321:
Ah ok. I'll see what I can do!

nicolasarnold12321:
@dtuite I added some resources to the documentation there

dtuite (approved):

PR #482, opened 2021-11-22
Add FAQ page and add section about status page

nicolasarnold12321:
https://deploy-preview-482--roadie.netlify.app/faqs

### DIFF
--- a/src/components/landing/faqs.js
+++ b/src/components/landing/faqs.js
@@ -29,14 +29,17 @@ const FAQ_CONTENT = [{
   id: 5,
   question: "Do you have a status page?",
   answer: () => {
-    return (
-      <p>
+    // this looks not nice but the linter was complaining that the displayname was not set
+    function statusText(){
+      return (<p>
         Yes we do! It can be found here:{' '}
         <a href='https://status.roadie.io'>
          https://status.roadie.io/
         </a>
-      </p>
-    )
+      </p>)
+    }
+
+    return statusText();
   },
 }];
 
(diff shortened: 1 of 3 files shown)


src/components/landing/faqs.js
```diff
@@ -8,29 +8,35 @@ const FAQ_CONTENT = [{
...
+  answer:
+    "Yes we do! It can be found here: <a href='https://status.roadie.io'>https://status.roadie.io/</a>",
 }];
 
 
 const Item = ({ question, answer }) => (
   <div>
     <dt className="font-semibold text-gray-900">{question}</dt>
-    <dd className="mt-3 text-gray-500">{answer}</dd>
+    <dd className="mt-3 text-gray-500" dangerouslySetInnerHTML={{ __html: answer }}/>
```
nicolasarnold12321, line 39: Sounds scary but i think in this case, its not 'dangerous' as we are only displaying simple text

Xantier: if you make the answer a function, React should be able to render is as JSX.

nicolasarnold12321: Ok cool thanks!

nicolasarnold12321: Done

Xantier: you can do something like 
```
{answer instanceOf Function ? asnwer() : answer}
```
so there's no need to necessarily wrap non-html to be html.

Xantier (approved):

Xantier (approved):
aight let's ship it

PR #513, opened 2021-12-09
Add GitHub holding page

sblausten:
Replacing this:

Screenshot 2021-12-09 at 11 10 58

with this:

Screenshot 2021-12-09 at 17 14 04


Xantier:
I think the message looks grand. Are we able to get rid of the manual step completely? We should be able to add additional state to the admin approval link or make it call a webhook or something right?

dtuite:
@sblausten How will they know when we have done our work? I don't think we are making that obvious to them.

sblausten:
> @sblausten How will they know when we have done our work? I don't think we are making that obvious to them.

Yes was thinking of that - doesn't seem we have been telling them up till now that it is installed. Ideally we'd email them as part of the manual steps. I can add that and then say in the message that we'll let them know via their github email address?

dtuite:
@sblausten I'd suggest saying we will email them is in scope for this PR. We can do that manually. The automation of that process is out of scope and can be done separately.

dtuite (approved):

PR #522, opened 2021-12-26
Dark mode

dtuite:
https://app.shortcut.com/larder/story/5635/dark-mode


martina-if:
Mi 2cents about the dark mode:

I don't personally use dark mode anymore but when I did it I wanted it on windows or applications where I spent a long time reading so IMO this makes sense on apps like github, IDEs, etc. Given this, if this is difficult, I would focus on adding support for it on the documentation and blog post pages which are where users might spent more time. At the same time, IMHO this is not super important on the marketing site.

When looking at other tech companies websites I see most are in light mode, but interestingly I found this: https://userstyles.org/styles/167671/kubernetes-io-dark-theme

PR #539, opened 2022-01-07
Generate tech docs for onboarding documentation

nicolasarnold12321:
PR for automatically generating roadie tech docs from marketing site docs.

Here is a link to them working: https://roadie.roadie.so/catalog/default/component/roadie-techdocs/docs/

[sc5694]


dtuite:
Can we control the order of the sidebar items in Backstage? They're supposed to be in a rough order, from most basic to most advanced.

Screenshot 2022-01-10 at 16 04 25

The statement on this page is not correct. Can we just remove the page entirely? There already is a whole section about plugins in the docs.

Screenshot 2022-01-10 at 16 06 11

The index page should be [this page](https://roadie.roadie.so/catalog/default/component/roadie-techdocs/docs/getting-started/getting-started-for-admins/) for the moment. Ideally it would be a less admin specific page but it's the best we have for now.

Screenshot 2022-01-10 at 16 07 45

nicolasarnold12321:
> Can we control the order of the sidebar items in Backstage? They're supposed to be in a rough order, from most basic to most advanced.

yes but the tech docs nav page would not be dynamically created

> The statement on this page is not correct. Can we just remove the page entirely? There already is a whole section about plugins in the docs.

sure

> The index page should be this page for the moment. Ideally it would be a less admin specific page but it's the best we have for now.

I can try figure something out

Xantier:
> yes but the tech docs nav page would not be dynamically created

We can introduce a naming convention for folders/files that would take care of this. `/1-getting-started.md`, `2-something-else.md` etc.

dtuite:
> I can try figure something out

It it's not possible, the index page can just link to the getting started page.

I like Jussi's idea to create and order for the sidebar. It's going to break the marketing site URLs though. We'd need to set up a lot of 301 redirects to maintain backwards compatibility.

The docs sidebar in the marketing site is manually created too, which is not ideal. Perhaps we can create a yaml representation of the sidebar and use it to drive both the marketing site and techdocs.

nicolasarnold12321:
I have added the index.md back in that now points to the adding component docs.

martina-if (commented):
Looks pretty neat!

I have a suggestion to add to the github actions workflow: in PRs add a step that checks that if someone modifies one of the doc files and (for any reason) doen't ´git add´ the generated docs it fails the checks. Something like:

´´´
yarn tech-docs:build
if ! git diff roadie-techdocs --exit-code; then
  echo "tech-docs not generated"
  git diff roadie-techdocs
  echo '::set-output .... fail this action'
fi
´´´

dtuite:
@nicolasarnold12321 For the sidebar issue,

We could probably quite easily create a YAML file which describes the order and contents of the sidebar in the marketing site. It would look something like this (truncated)

```yaml
docs:
  section: 'Getting started'
    - 'Configuring Roadie': '/docs/getting-started/getting-started-for-admins/'
    - 'Adding components': '/docs/getting-started/adding-components/'
  section: 'Integrations'
    - 'ArgoCD': '/docs/integrations/argocd/'
    - 'Bugsnag': '/docs/integrations/bugsnag/'
    - 'Buildkite': '/docs/integrations/buildkite/'
    - 'CircleCI': '/docs/integrations/circleci/'
```

That YAML file could be used to generate both `src/components/doc/Sidebar.js` and the sidebar in MkDocs.

Do you think this could be a way to solve the sidebar ordering problem?

nicolasarnold12321:
> That YAML file could be used to generate both

It certainly can! This will not make the techdocs generation dynamic though. Is that something that we would like to consider?

martina-if:
sc5694

martina-if:
https://app.shortcut.com/larder/story/5694/understand-how-to-send-roadie-docs-to-marketing-site-and-backstage

nicolasarnold12321:
@martina-if I have added the validation to the github action.

roadie-tech-docs/docs/configuration/tech-radar/index.md
```diff
@@ -0,0 +1,111 @@
+
+## Introduction
```
dtuite, line 2: Could we put all this duplicate content into a different GitHub repo? I'd like to have separation between the content which is the source of truth and the content which is just a result of the build process.

static/docs-nav.yaml
```diff
@@ -0,0 +1,42 @@
+site_name: 'Roadie Onboarding docs'
```
dtuite, line 1: The static directory is not the right place for this file. Stuff in the static dir is copied into the `public` dir when Gatsby builds so this file will be publically available at https://roadie.io/static/docs-nav.yaml

I would put this in `src`.

dtuite (changes requested):

dtuite:
@nicolasarnold12321 when people write docs in the future, they are likely to paste in links that start with `https://roadie.io/...`. We've already seen examples of this happening since you've had to clean them up. This is a problem because we won't know that something is wrong until a customer clicks the link and is taken to roadie.io. Even then, they probably won't report it.

Can we make the build fail please if it detects a URL in the `content/docs` directory which starts with `https://roadie.io`? That way people will be notified if they accidentally use the wrong URL. You could probably accomplish this with a `grep` in the GitHub Action.

static/docs-nav.yaml
nicolasarnold12321: I put it in content to match where these docs reside

roadie-tech-docs/docs/configuration/tech-radar/index.md
nicolasarnold12321: Done

dtuite (approved):

PR #542, opened 2022-01-11
Update jira.md

belaljuma:
two proposed changes. The first is to explicitly ensure the addition of the package is in the right place (packages/app), and the second is to surround the target JIRL URL with single quotes


dtuite:
Lovely! Thank you.

dtuite (approved):

PR #550, opened 2022-01-17
Add open roles to About page

dtuite:
https://app.shortcut.com/larder/story/6029/add-open-roles-to-about-page

### DIFF
--- a/src/components/RecruiteeJobsList.js
+++ b/src/components/RecruiteeJobsList.js
@@ -35,7 +35,7 @@ const RecruiteeJobsList = () => {
     return () => {
       document.body.removeChild(rtscript);
     }
-  });
+  }, []);
 
   return (
     <div className="bg-white">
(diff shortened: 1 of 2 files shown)


src/components/RecruiteeJobsList.js
```diff
@@ -0,0 +1,61 @@
...
+        }
+      });
+    }
+    rtscript.src = 'https://d10zminp1cyta8.cloudfront.net/widget.js';
+    document.body.appendChild(rtscript);
+
+    return () => {
+      document.body.removeChild(rtscript);
+    }
+  });
```
Xantier, line 38: Add deps array
```suggestion
  }, []);
```

dtuite: Interesting. I kinda understand what that does when there are items in the array. What's the difference between an empty array and none at all?

Xantier: none at all will rerun the function on every render. Empty on the first render. It doesn't make a difference on this site because the page will be rendered most likely once (Gatsby prerenders and it'll be HTML already), but on other sites it may cause an infinite loop or dupe resources etc.

dtuite: Appreciate the explanation!

Xantier (approved):

PR #590, opened 2022-02-15
Add explanation how to customize graphs in TechDocs

Irma12:
As discussed in standup, we decided to include small explanation around how graphs can be customisable.

Preview: https://deploy-preview-590--roadie.netlify.app/docs/getting-started/technical-documentation/#customize-graphs-in-techdocs


content/docs/getting-started/technical-documentation/index.md
```diff
@@ -93,6 +94,13 @@ On the Overview page, click Docs in the tab bar. You should now see your documen
 
 ![a page of basic documentation for a service in the Backstage service catalog](./sample-service-docs-backstage.png)
 
+## Customize graphs in TechDocs
+
+When using mkdocs with TechDocs, base used for defaults such as plugins, theming etc are contained in [mkdocs-techdocs-core](https://github.com/backstage/mkdocs-techdocs-core)
+plugin. This means markdown-inline-graphviz extension is being used to replace Graphviz definitions with inline SVGs or PNGs. As long as `graphviz` is available in the environment
```
dtuite, line 100: This part about the environment is not useful for the reader. They have no control over the environment, we control that.

content/docs/getting-started/technical-documentation/index.md
```diff
@@ -93,6 +94,13 @@ On the Overview page, click Docs in the tab bar. You should now see your documen
...
+where the docs are generated this will work out of the box, but in order to customize the look you need to use the [Graphviz attributes](https://graphviz.org/doc/info/attrs.html).
```
dtuite, line 101: This last small sentence is actually the important part. Can we show an example of how to change the contrast please? That was the query that triggered this work in the first place.

Irma12: I understand the point here but I was looking at Graphviz documentation regarding different attributes which could be used to customize the look and didn’t want to end up duplicating their documentation. I could include background color change or other set of attributes you think could be useful

dtuite: You're definitely right that we don't want to duplicate too much of the docs here because it will get stale too quickly. At the same time, there is a balance between duplicating the docs, and making the reader click the link and read a whole new set of docs in order to answer their question (which is basically... "how do I flip the colors on this diagram").

I haven't read the graphviz docs myself, but if there's a small snippet that you can duplicate here in order to give the reader a sense of how to change the background color or something simple like that, then I think it's worth doing.

dtuite (approved):

content/docs/getting-started/technical-documentation/index.md
Irma12: @dtuite do you think this way makes it bit better?

PR #591, opened 2022-02-15
Clear forms after submit

dtuite:
This won't stop bot spam but it at least stops people from sitting on the form and resubmitting over and over with 2 clicks each time.


dtuite:
https://app.shortcut.com/larder/story/6785/prevent-rapid-form-submissions

PR #608, opened 2022-03-04
Add pro pricing plan

dtuite:
Preview URL: https://deploy-preview-608--roadie.netlify.app/pricing/


dtuite:
https://app.shortcut.com/larder/story/7868/add-a-pro-pricing-plan

src/components/pricing/ProPricingTier.js
```diff
@@ -0,0 +1,100 @@
...
+        </div>
+      </div>
+
+      <TierBulletsSection
+        hasIcon={false}
+        bullets={[
+          'Unlimited software components tracked.',
+          'Unlimited API specs.',
+          'Unlimited TechDocs',
+          'Unlimited scaffolder templates.',
```
Xantier, line 77: Did we want to put some notion of public Discord/Slack support in here?

dtuite: Not yet. I can add that after I set up some sort of forum or something.

src/components/landing/faqs.js
```diff
@@ -1,46 +1,39 @@
 import React from 'react';
-import { Headline } from 'components';
+import { Headline, TextLink as Link } from 'components';
 
 const FAQ_CONTENT = [{
-  id: 1,
   question: "Is Roadie SaaS or on-prem?",
   answer: () => 
     "Roadie is SaaS only for the moment. Some Backstage plugins will need a connection back to your infrastructure. We offer a number of secure methods for enabling this.",
```
Xantier, line 7: Just saw this on the preview. "SaaS only for the moment" makes me think that it's only currently SaaS but won't be in the future. Maybe we can flip the words 'only' and 'SaaS'

src/components/pricing/GrowthPricingTier.js
```diff
@@ -0,0 +1,60 @@
...
+const GrowthPricingTier = () => (
+  <div className="border border-gray-200 rounded-lg shadow-sm divide-y divide-gray-200">
+    <div className="p-6">
+      <TierName name="Growth" />
+      <TierDescription description="For excellent engineering orgs who want to maintain effectiveness through hypergrowth" />
+
+      <p className="text-4xl font-extrabold text-gray-900 mb-14">Custom</p>
+
+      <div className="mb-10">
+        <TextField
```
Xantier, line 17: Does this necessarily need to be a text field if they will be disclosing numbers in the demo anyways. 
Also are we alienating customers that want SLA and support etc. but don't have an org worth of 150 seats yet when we use the text '150 developers or more'?

src/components/landing/faqs.js
dtuite: What if I just remove "for the moment"? Would that make more sense?

src/components/pricing/GrowthPricingTier.js
dtuite: Let's continue this conversation on the clubhouse ticket.

Xantier (approved):

PR #639, opened 2022-03-21
Split Initial configuration doc

dtuite:
Having two items under "Getting started" will make it easier for people to find the specific doc they need. "Initial configuration" means a lot less than "Install GitHub app".

Preview link: https://deploy-preview-639--roadie.netlify.app/docs/getting-started/install-github-app/


dtuite:
https://app.shortcut.com/larder/story/7780/split-initial-configuration-doc

content/docs/getting-started/install-github-app/index.md
```diff
@@ -0,0 +1,43 @@
...
+description: How to give Roadie the access it needs to your GitHub org.
+---
+
+## Introduction
+
+This tutorial will guide you through the steps required to connect Roadie to your GitHub org.
+
+## Prerequisites
+
+If your IT department filters your outbound access by hostname, you may need to allow access to Roadie hostnames before proceeding. More information on this can be found [here](/docs/details/allowlisting-roadie-traffic/).
```
Xantier, line 13: I think the allow list page doesn't list hostnames, just our IPs. We should add those hostnames in there as well though.

dtuite: Ok. I don't know what the hell I did in https://github.com/RoadieHQ/marketing-site/pull/625 but somehow I removed the hostnames completely from the allowlist docs.

dtuite: This PR doesn't cause that problem though. So can I merge this one and fix the allowlist in a separate PR today?

Xantier (approved):

content/docs/getting-started/install-github-app/index.md
Xantier: :+1:

content/docs/docs-nav.yaml
```diff
@@ -6,23 +6,24 @@ plugins:
 
 nav:
   - Getting started:
-    - Initial configuration: '/docs/getting-started/getting-started-for-admins/'
+    - Installing the GitHub app: '/docs/getting-started/install-github-app/'
+    - Designating admins: '/docs/getting-started/create-admin-group/'
```
punkle, line 10: I wonder should this be in its own section with inviting users and called it "User Management"?

content/docs/getting-started/create-admin-group/index.md
```diff
@@ -0,0 +1,42 @@
...
+
+5. Ensure you are logged in to GitHub from Roadie. Click the link in the sidebar to do this.
+
+![A link that says "Log in to GitHub"](./sidebar-log-into-github.png)
+
+⚠️  &nbsp;It can take some time for Roadie to refresh the list of teams from GitHub teams. If you do not see admin functions immediately, please wait a few minutes and try again.
+
+## Next steps
+
+Now that you have connected to GitHub and made yourself an admin, you can [start adding components](/docs/getting-started/adding-components/).
```
punkle, line 42: Should we have a link to this page from the inviting a user page?

content/docs/docs-nav.yaml
dtuite: The idea is that the Getting Started docs are in order. My thought was that after installing the GitHub app, making an admin is the next most important thing for the user to do.

However, now that I think about it, if they can add components etc. without making themselves an admin (because they are an admin by default), then you are right. The instructions for making admins are not useful until they want to add a new user.

Is that what you're thinking?

content/docs/getting-started/install-github-app/index.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Installing the GitHub App
+publishedDate: '2022-03-21T21:00:00.0Z'
+description: How to give Roadie the access it needs to your GitHub org.
+---
+
+## Introduction
+
+This tutorial will guide you through the steps required to connect Roadie to your GitHub org.
```
punkle, line 9: Should we have a page that gives you a bit of orientation.

I feel like this page should be called something like: "Configuring GitHub as a catalog source" or something like that, then when Jussi is done, it'd have a similar page for Bitbucket.

Also it feels like this intro could say something like:

"This tutorial will guide you through the steps required to configure Backstage to index your GitHub Organization projects and teams by installing a GitHub App."

content/docs/docs-nav.yaml
punkle: Yes.

dtuite: Yeah good idea. I've done the following...

1. Move "Designating admins" after "Inviting users" in the sidebar. This should mean they can invite some users who will all be admins and then learn how to decide who is and isn't an admin afterwards.
2. Link to "Designating admins" in the Next Steps section of the "Inviting users doc".

Does that make more sense now?

content/docs/getting-started/install-github-app/index.md
dtuite: I was adding more to this PR to incorporate Brian's suggestions so I made this change while I was at it.

content/docs/getting-started/create-admin-group/index.md
dtuite: Yes. Done

content/docs/getting-started/install-github-app/index.md
dtuite: That's pretty much the direction I was going here. I wanted to break out the individual pages so I can link to them from a docs homepage. That will come in a different PR.

dtuite:
@punkle @Xantier I have addressed all comments.

content/docs/docs-nav.yaml
punkle: yes.

punkle (approved):

PR #640, opened 2022-03-23
Sc 7454 AWS S3 docs

iain-b:
Document configuring AWS S3... to do that I made a generic section about how to create roles for roadie based on the k8s section. I also moved all the plugins stuff to plugins/ rather than integrations/ as I felt S3 belonged in integrations (and we're likely to add other stuff there soon).


Xantier (dismissed):

Xantier (approved):

iain-b:
More changes after testing on our tenant.

content/docs/details/accessing-aws-resources/index.md
```diff
@@ -87,7 +89,7 @@ You should see a page like this
         },
         "StringLike": {
           "aws:PrincipalArn": [
-            "<ROADIE BACKEND ROLE ARN>"
+            "*<ROADIE BACKEND ROLE>*"
```
Xantier, line 92: Maybe could be worth mentioning why the starts (we autogenerate the ending of the role or something?)

Xantier: Like in the note below but an actual example

Xantier (approved):

PR #651, opened 2022-03-28
Docs tabs

dtuite:
https://app.shortcut.com/larder/story/7964/move-docs-categories-into-tabs

Preview: https://deploy-preview-651--roadie.netlify.app/docs/getting-started/install-github-app/

Full screen size:

Screenshot 2022-03-28 at 21 31 01

Mobile size:

Screenshot 2022-03-28 at 21 31 47

With the hamburger menu open:

Screenshot 2022-03-28 at 21 31 53


content/docs/docs-nav.yaml
```diff
@@ -47,9 +47,9 @@ nav:
...
     - Authenticated Proxy: '/docs/custom-plugins/authenticated-proxy/'
-  - Details:
-    - GitHub app permissions: '/docs/integrations/github-app-permissions/'
+  - In-depth:
+    - GtHub app permissions: '/docs/details/github-app-permissions/'
     - Setting secrets: '/docs/details/setting-secrets/'
-    - Updating the UI: '/docs/getting-started/updating-the-ui/'
+    - Updating the UI: '/docs/details/updating-the-ui/'
     - Allowlisting Roadie Traffic: '/docs/details/allowlisting-roadie-traffic/'
-    - Required annotations: '/docs/integrations/required-annotations'
```
dtuite, line 55: I'm changing these URLs so I can make a simple rule that all URLs with "integrations" in them can go into the "Plugins and integrations" tab and all URLs with "details" in them can go in the "In-depth" tab.

Redirects for the old URLs are handled in the netlify.toml below.

src/components/SitewideHeader/DocsDrawerMenu.js
```diff
@@ -0,0 +1,46 @@
...
+
+import DrawerMenuHeader from './DrawerMenuHeader';
+import MobileDropdownNavItem from './MobileDropdownNavItem';
+import { DOCS_LAYOUTS } from 'components/doc';
+
+const subItems = DOCS_LAYOUTS.map((props) => ({
+  name: props.tabLabel,
+  to: props.startPath,
+  isActiveMatch: props.isActiveMatch,
+}));
```
dtuite, line 12: The drawer menu for the docs now just shows the docs categories so it needs to be different than the drawer menu for the rest of the site.

src/components/legal-notice/Sidebar.js
```diff
@@ -1,54 +1,51 @@
...
-      />
-
-      <SidebarItem
-        to="/legal-notices/acceptable-use-policy/"
-        text="Acceptable Use Policy"
-        partiallyActive={true}
-      />
-    </SidebarSectionList>
+    <SidebarSection>
+      <ul className="p-0 list-none">
```
dtuite, line 11: The `<SidebarSectionList />` no longer takes `children` so we have to use a plain `<SidebarSection />`.

src/components/doc/layouts.js
```diff
@@ -0,0 +1,19 @@
...
+
+export const DOCS_LAYOUTS = [{
+  tabLabel: 'Getting started',
+  startPath: '/docs/getting-started/install-github-app/',
+  isActiveMatch: /\/docs\/getting-started/,
+  sidebarNavItemGroups: [sidebar.nav[0]],
+}, {
+  tabLabel: 'Plugins & Integrations',
+  startPath: '/docs/integrations/argocd/',
+  isActiveMatch: '/docs/(?:custom-plugins|integrations)/',
```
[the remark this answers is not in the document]
Xantier: Would it make sense to add a redirect to the root of `/docs/integrations` and `/docs/custom-plugins` (and others) to point to argocd instead of hardcoding the first `to` value here? 
Currently the path to those top level URLs returns 404.

Xantier:
I think we should make a distinction between 'plugins' and 'integrations' at least on the sidebar.
I think these 4 
image
should be in integrations with the upcoming S3 and Bitbucket docs

src/components/doc/layouts.js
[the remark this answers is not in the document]
dtuite: That would make the code look like this right? I wouldn't be able to get rid of the hardcoded value, but it would be a less specific hardcoded value tan it currently is.

```
tabLabel: 'Plugins & Integrations',
startPath: '/docs/integrations/',
isActiveMatch: '/docs/(?:custom-plugins|integrations)/',
```

I do agree that `/docs/integrations` and `/docs/custom-plugins` should not 404. I've made [this ticket](https://app.shortcut.com/larder/story/7997/make-an-integrations-homepage) to fix that. Until that ticket is delivered, perhaps we can live with `startPath: '/docs/integrations/argocd/',`?

dtuite:
> I think we should make a distinction between 'plugins' and 'integrations' at least on the sidebar. I think these 4 image should be in integrations with the upcoming S3 and Bitbucket docs

I agree. I've made https://app.shortcut.com/larder/story/7999/separate-out-integrations for that.

Xantier (approved):

PR #656, opened 2022-03-29
Sc 7686 gcalendar docs

iain-b:
Preview: https://deploy-preview-656--roadie.netlify.app/docs/integrations/gcalendar/


dtuite:
@iain-b seems like this needs a rebase or something. The S3 docs are in here too.

content/docs/integrations/gcalendar/index.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Google Calendar Plugin
```
dtuite, line 2: Should this be "Home Page Google Calendar" to match "Home Page Markdown"?

I'm just thinking it's a good idea to have these home page widgets collected together in the sidebar.

content/docs/integrations/gcalendar/index.md
```diff
@@ -0,0 +1,27 @@
...
+
+The Google Calendar plugin provides a home page card you can use to display your agenda in Backstage.
+
+![gcp-api-search](./gcalendar-card.png)
+
+## Installation
+
+1. Set up google auth in Roadie following [this guide](/docs/integrations/google-oauth-client)
+
+2. Ensure the Google calendar API is enabled in GCP.
```
dtuite, line 17: I think it's worth moving these into headings rather than having a nested numbered list.

I can see it working like:

h2 Set up Google OAuth

Set up google auth in Roadie following [this guide](/docs/integrations/google-oauth-client)

h2. Add the Google Calendar card

Steps 1, 2 and 3.

content/docs/integrations/gcalendar/index.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Introduction
+
+The Google Calendar plugin provides a home page card you can use to display your agenda in Backstage.
+
+![gcp-api-search](./gcalendar-card.png)
+
+## Installation
+
+1. Set up google auth in Roadie following [this guide](/docs/integrations/google-oauth-client)
```
dtuite, line 15: Google should be capitalized.

content/docs/integrations/gcalendar/index.md
iain-b: While currently it only supplies a home page card the review when it was introduced  asked that it be named generically so any future calendar widgets could be added here (I think even gsuite was suggested). We probably should have done that with the home page markdown plugin too.

content/docs/integrations/gcalendar/index.md
iain-b: Yep, I'll do that.

content/docs/integrations/gcalendar/index.md
iain-b: This doesn't stop us linking to the "add google calendar to the home page" from a getting started page around the home page for example or from next steps in the current home page docs

dtuite (approved):

PR #682, opened 2022-04-21
Update jenkins.md

franzqat:
The square brackets may lead into a mislead. The value is supposed to be a string.
Also the instance name can be included and the organization is optional as well.


dtuite:
@franzqat Thanks for the contribution.

Unfortunately it seems that our builds won't pass for non-member contributions. I will integrate this change into an upcoming PR instead.

Xantier (approved):
Closing as part of https://github.com/RoadieHQ/marketing-site/pull/767

PR #687, opened 2022-04-25
Allow twitter and YT embeds

dtuite:
https://app.shortcut.com/larder/story/6782/add-csp-to-roadie-io


dtuite:
Here's an example of a broken page: https://roadie.io/blog/backstage-weekly-5-lighthouse/

dtuite:
@martina-if merged this without +1 to fix some blog posts which would have had broken embeds.

PR #701, opened 2022-04-29
fix: memberOf should be `spec.memberOf` for User Entities

knksmith57:
When following the [_Manual admin group configuration with group entities (advanced alternative)_ guide](https://roadie.io/docs/getting-started/create-admin-group/#manual-admin-group-configuration-with-group-entities-advanced-alternative), the yaml snippet for the example User Entity has `memberOf` at the root —it should be at `spec.memberOf` instead.


dtuite:
Tackled in #688

PR #722, opened 2022-05-16
revert getsiteurl functionality

dtuite:


dtuite:
Merging without review to fix RSS production bug.

PR #728, opened 2022-05-17
add outputs and explain contacting third party services

punkle:


content/docs/details/scaffolder/index.md
```diff
@@ -232,6 +232,43 @@ If you would like to prompt the user to add entity tags, you can use the `ui:fie
         type: array
         ui:field: EntityTagsPicker
 ```
+### Outputs
+Parameters can be retrieved later on by steps using parameter outputs. Here is an example of a parameter `name` be used by a `debug:log` step.
```
Xantier, line 236:
```suggestion
Parameters can be retrieved later on by steps using parameter outputs. Here is an example of a parameter `name` being used by a `debug:log` step.
```

content/docs/details/scaffolder/index.md
```diff
@@ -232,6 +232,43 @@ If you would like to prompt the user to add entity tags, you can use the `ui:fie
...
+         type: string
+   steps:
+    - id: log-message
+      name: Log Message
+      action: debug:log
+      input:
+        message: 'Hello, ${{ parameters.name }}!'
+```
+
+If you need to referene elements of an array parameter you can refer to them using the following syntax:
```
Xantier, line 251:
```suggestion
If you need to reference elements of an array parameter you can refer to them using the following syntax:
```

content/docs/details/scaffolder/index.md
```diff
@@ -505,6 +547,19 @@ steps:
         - java
         - ruby
 ```
+#### Outputs
+The `publish:github` action produces two step outputs: `remoteUrl` is the url for the newly created repository; `repoContentsUrl` is a url that shows the contents of the repository.
```
Xantier, line 551: I'd say these are better as a list so they can be scanned by the reader instead of having to read through a horizontal text

content/docs/details/scaffolder/index.md
```diff
@@ -866,11 +957,37 @@ steps:
...
+The `debug:log` action does not have any outputs.
+
 ### Other Actions
 You can find all of the actions available to your Backstage instance by visiting the following page from within Backstage:
 
 `https://<tenant-name>.roadie.so/create/actions`
 
+## Advanced
+### Calling an internal API
+If you need a scaffolder step to contact a custom authenticated service or any public API for that matter that is not currently supported by a built in action, you can do that using a combination of the `http:backstage:request` action and a [backstage proxy configuration](/docs/custom-plugins/authenticated-proxy/).
```
Xantier, line 970: This doesn't necessarily need to be an authenticated proxy. We should probably surface the simpler use case of using a standard proxy with a secret first

punkle: That's true. I'll do that this morning.

content/docs/details/scaffolder/index.md
punkle: Good idea

content/docs/details/scaffolder/index.md
punkle: If you dont mind, Ill follow through the customer secrets and proxy configuration changes in a different commit.

Xantier (approved):

PR #729, opened 2022-05-17
Troubleshooting catalog docs

iain-b:


content/docs/details/troubleshooting-the-catalog/index.md
```diff
@@ -0,0 +1,66 @@
+---
+title: Troubleshooting the Catalog
+publishedDate: '2022-05-17T21:00:00.0Z'
+description: How to resolve issues getting data into the catalog
+---
+
+## About Locations
+
+The backstage catalog uses location entities to point to external sources (in version control) which defined entities.
```
Xantier, line 9:
```suggestion
The backstage catalog uses location entities to point to external sources which define entities.
```
Not necessarily SCM, we have S3 enabled as well right?

Xantier (approved):

content/docs/details/troubleshooting-the-catalog/index.md
dtuite, line 9: Could we go into more depth here perhaps? What does a location entity look like? Like I've been using B backstage for 2 years and I'm honestly still confused about locations and how they get interpreted by Backstage.

Xantier: I'll continue on top of these docs and add more stuff regarding location mgmt etc.

Xantier:
Merging in Iain's absence

PR #758, opened 2022-06-06
Edits jun

dtuite:
https://app.shortcut.com/larder/story/9203/simplify-homepage


src/pages/index.js
```diff
@@ -51,31 +40,27 @@ const SEO_TITLE = 'The easiest way to get Spotify Backstage';
 const LEAD = `Roadie's SaaS platform handles hosting and upgrades and ensures you always have access to the latest Backstage features.`;
 
 const PRODUCT = {
-  htmlId: 'product',
-  title: 'Spotify Backstage, with benefits...',
-  description: `SaaS Backstage is simple to set up and maintenance free.`,
   features: [{
     title: 'Quick and easy setup',
-    description: 'Customize Backstage in a couple of clicks. Drag plugins to where you want them. Add new ones instantly.',
+    description: 'Add plugins in a couple of clicks. Drag widgets to the place you want them.',
```
dtuite, line 45: Snappier copy across the board

src/pages/installation-pending.js
```diff
@@ -15,13 +15,18 @@ You can also check in on the progress by contacting us at support@roadie.io or v
 const GithubAppInstallHoldingPage = () => (
   <>
     <SitewideHeader />
+
+    <div className="bg-white">
```
dtuite, line 19: Have to update these since I extracted the background color from the inside of the component.

src/components/landing/FeatureBlocks/AlternatingFeatureBlock.js
```diff
@@ -0,0 +1,109 @@
...
+        classnames('mt-10 -mx-4 relative lg:mt-0', { 'lg:col-start-1': illustrationSide === 'left' })
+      }
+      aria-hidden={illustrationSide === 'right'}
+    >
+      <DotPattern
+        className="absolute left-1/2 transform -translate-x-1/2 translate-y-16 lg:hidden"
+        width={784}
+        height={404}
+        aria-hidden={illustrationSide === 'left'}
+        id="ca9667ae-9f92-4be7-abcb-9e3d727f2941"
```
Xantier, line 49: This will possibly introduce multiples of the same ids to the site if there are many of these components used. Better to derive it from something instead. (Or drop it if it's not used anywhere). There are few more of these below also

src/pages/index.js
```diff
@@ -149,30 +100,53 @@ const SOLUTIONS = {
...
+      description: 'Spotify experienced an explosion of docs writing after adopting this methodology internally.',
+      icon: PresentationChartLineIcon,
+    }],
   }],
 };
 
 const PLUGINS_SUPPORTED = {
-  title: 'Pre-loaded with the best plugins',
-  subTitle: `Open-source plugins are supported by default. Push your own plugins if you need something bespoke.`,
+  title: 'Powerful plugins',
```
Xantier, line 148: We should probably increase the amount of plugins displayed here. There are a bunch more logos we can add.

Xantier (dismissed):

src/components/landing/FeatureBlocks/AlternatingFeatureBlock.js
dtuite: I'm gonna skip over this comment since I'm just moving this code rather than adding this.

But you are right that something is not good here.

Xantier (approved):

PR #763, opened 2022-06-13
Add docs on escape syntax and preview tool

sblausten:
https://app.shortcut.com/larder/story/9282/yotpo-relativity-escaping-variables-in-scaffolder-template


content/docs/details/scaffolder/index.md
```diff
@@ -1091,10 +1095,18 @@ steps:
...
 
+### Escaping syntax
+
+If you need to pass variable substitution syntax through without it being interpreted you can escape the syntax by wrapping it like so `${{ '${{ parameters.something }}' }}`.
+
 ## Troubleshooting
 
 Writing templates can be a little cumbersome at times. We have compiled a list of errors that we have seen in the past, that might help you determine the cause of your issue.
 
+Template YAML input forms can also be tested at `/administration/tools` using a live template preview viewer which speeds up the testing cycle.
```
Xantier, line 1106: I think there's no need to mention this twice (especially with the image) since it's added on the top already.

sblausten: I think with a doc this big it is probably a good idea - as far as I know its best practice to repeat info throughout docs as there are many different routes users can come from and its easy to miss if only in a single place.

dtuite (approved):

PR #766, opened 2022-06-14
Adds documentation for Vault

jorgelainfiesta:
## Motivation

Last week, @ivangonzalezacuna released a plugin that integrates Vault with Backstage. We want to keep our docs helpful by including its documentation. 

## Approach 

I tried the plugin on a local Backstage instance, following the installation instructions and they worked just fine. I went through a lot of problems trying to make Vault work, but that's on me for I had no experience with it. 

[Preview link](https://deploy-preview-766--roadie.netlify.app/backstage/plugins/hashicorp-vault/)


dtuite (approved):
Perfect!

PR #779, opened 2022-06-30
Add docs to explain Confluence search integration

Xantier:


content/docs/integrations/confluence/index.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Store Token as a Secret to Roadie
+
+When you have successfully encoded your token, you can navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/secrets` and populate the secret value `CONFLUENCE_TOKEN`.
+
+![Confluence Secret Token Screen](./confluence_secret.png)
+
+### Configure Confluence Search Indexing
+
+Final step to enable the Confluence wiki for search indexing is to configure your Confluence organization name, and the cadence you want your Confluence documents to be indexed. You can navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/confluence` to input the needed values.
```
martina-if, line 61: Could be nice to mention min and max allowed values

content/docs/integrations/confluence/index.md
```diff
@@ -0,0 +1,69 @@
...
+You can create an Atlassian token by navigating to your [**Atlassian** profile security page](https://id.atlassian.com/manage-profile/security) and Clicking 'Create and Manage API Tokens'.
+![Creating an Atlassian token](./confluence_api_token_generation.png)
+
+
+On the page navigated into you are able to generate API tokens which can be used to connect Roadie to your Confluence workspaces.
+
+
+### Construct Confluence Authentication Token to Work with Roadie
+
+The Confluence API uses Basic Authentication on their API to negotiate authentication. To create a compatible token you need to **base64 encode** your username and API token pair. To achieve that you can either use any of the online encoders or a command line tool.
```
martina-if, line 34: I think we had a similar pattern with the jira (?) plugin and it didn't work well to ask the user to base64 encode things. No need to change this now but would be worth considering that when we need this we should ask for the user and password separately and do the base64 encoding in our logic. In this case it's a little cumbersome because we need to ask for the secret in one place and the user in a different one 🤔

martina-if (approved):

content/docs/integrations/confluence/index.md
Xantier: :+1:

content/docs/integrations/confluence/index.md
Xantier: Good idea :+1: Changed, and modified the implementation to match as well

content/docs/integrations/confluence/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Store Token as a Secret to Roadie
+
+When you have successfully generated your token, you can navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/secrets` and populate the secret value `CONFLUENCE_TOKEN`.
+
+![Confluence Secret Token Screen](./confluence_secret.png)
+
+### Configure Confluence Search Indexing
+
+Final step to enable the Confluence wiki for search indexing is to configure your Confluence organization name, username tied to the token and the cadence you want your Confluence documents to be indexed. You can navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/confluence` to input the needed values. The minimum interval is 15 minutes to prevent excessive load on the used API token and Confluence API.
```
jorgelainfiesta, line 39:
```suggestion
To enable your Confluence wiki for search indexing,  navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/confluence` and input your organization name, username, Confluence API token, and indexing interval. The fastest interval rate we support is scheduling indexes every 15 minutes to prevent excessive use of Confluence's API token.
```

"Minimum of 15 minutes" is a bit confusing IMO (also in the UI). The meaning of the question is a bit ambiguous.

content/docs/integrations/confluence/index.md
```diff
@@ -0,0 +1,47 @@
...
+When you have successfully generated your token, you can navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/secrets` and populate the secret value `CONFLUENCE_TOKEN`.
+
+![Confluence Secret Token Screen](./confluence_secret.png)
+
+### Configure Confluence Search Indexing
+
+Final step to enable the Confluence wiki for search indexing is to configure your Confluence organization name, username tied to the token and the cadence you want your Confluence documents to be indexed. You can navigate to `https://<your-roadie-instance>.roadie.so/administration/settings/confluence` to input the needed values. The minimum interval is 15 minutes to prevent excessive load on the used API token and Confluence API. 
+![Confluence Search Indexing Configuration](./confluence_settings.png)
+
+After clicking Apply & Restart, the search engine on Roadie should start populating the search index with Confluence documents. Not that the initial retrieval of wiki pages may take some time to kick off and populate, depending on the amount of documents in Confluence.
```
jorgelainfiesta, line 42:
```suggestion
After clicking "Apply & Restart", the search engine on Roadie will start populating the search index with Confluence documents. Note that the first time we retrieve your Wiki may take some time to populate, depending on the number of documents in Confluence.
```

jorgelainfiesta (commented):
Great work! left some suggestions for your consideration.

content/docs/integrations/confluence/index.md
```diff
@@ -0,0 +1,47 @@
...
+---
+
+## Introduction
+
+Roadie supports integration with Atlassian Confluence wiki platform to enable cross application search functionality. Roadie indexes documents from Confluence so they are available for you to search within the Roadie search bar. 
+
+
+## Prerequisites
+
+You'll need an Atlassian account with an ability to create an API token and knowledge of the name of your Confluence organization.
```
Xantier, line 18:
```suggestion
You'll need an Atlassian account with an ability to create an API token and the name of your Confluence organization.
```

PR #782, opened 2022-07-05
New landing page with CTA

jorgelainfiesta:
## Approach 

All of our traffic at the moment comes from a Backstage context, so I propose we optimize for this audience. I'm suggesting we try an inline form to see how that performs. We can revert back to buttons if it doesn't work well enough. 

## Screenshot

Screenshot 2022-07-20 at 10 53 00


src/components/landing/CenteredHero/HeroForm.js
```diff
@@ -0,0 +1,191 @@
...
+    provided.
+  </p>
+);
+
+const SubmissionSuccessNegativeBody = () => (
+  <>
+    <p>Roadie only supports GitHub for now.</p>
+    <p>
+      We are working to support more tools in the near future. You will be among the first to know
+      when we support yours.
```
dtuite, line 57: One thing that has happened before is that I received an angry email from a person who submitted this form after picking "Bitbucket", and were then told that we don't support bitbucket.

They felt that they had been tricked into supplying their email address, even though we didn't support their tool.

If we can prevent people feeling this way without making the form too ugly I think it would be beneficial.

src/components/landing/CenteredHero/HeroForm.js
```diff
@@ -0,0 +1,191 @@
...
+            fullWidth={true}
+            text={buttonText}
+            disabled={disabled}
+            required={true}
+            className="font-bold tracking-wide"
+          />
+        </div>
+
+        <div className="sm:col-span-4">
+          <p className="text-sm text-white opacity-60">
```
dtuite, line 160: If this legal stuff is duplicated on the other form, it would be worth extracting it into a component.

src/environment.js
```diff
@@ -18,3 +18,4 @@ export const chargebeeSubdomain = () => {
 };
 
 export const siteRecaptchaKey = () => process.env.GATSBY_SITE_RECAPTCHA_KEY;
+export const siteRecaptchaInvisibleKey = () => process.env.GATSBY_SITE_RECAPTCHA_INIVISIBLE_KEY;
```
dtuite, line 21: Does this actually work? I thought Netlify didn't support it.

src/pages/index.js
```diff
@@ -16,19 +16,17 @@ import {
   ChartBarIcon,
   BadgeCheckIcon,
   CubeTransparentIcon,
-} from '@heroicons/react/outline'
+} from '@heroicons/react/outline';
```
dtuite, line 19: Gotta get them semi-colons! ❤️

src/pages/index.js
```diff
@@ -45,121 +43,157 @@ const SEO_TITLE = 'The easiest way to get Spotify Backstage';
...
+      ],
+    },
+    {
+      title: 'Keep your Backstage ever-green',
       description:
-        `We regularly patch vulnerabilities in the open-source code, ensuring you're using a hardened version of Backstage.`,
-      icon: ShieldCheckIcon,
-    }],
-  }],
+        'All Roadie customers are running on the latest Backstage version (1.3.0). Get the latest features from the community without the maintenance burden.',
```
dtuite, line 74: This version is already out of date. I think it will be quite burdensome to have to remember to update this all the time.

src/pages/index.js
```diff
@@ -45,121 +43,157 @@ const SEO_TITLE = 'The easiest way to get Spotify Backstage';
...
+      },
+      bullets: [
+        {
+          name: 'Automated upgrades',
+          description: `Self-hosted Backstage teams can take up to 30% of their time upgrading their instance. We manage upgrades for you, whether they're security patches, minor, or major releases.`,
+          icon: ChipIcon,
+        },
+        {
+          name: 'SLAs Available',
+          description: `Downtime in Backstage can cause friction in your team and hinder adoption. We offer reliability and 24/7 support.`,
```
dtuite, line 88: Careful with "24/7 support". We have 24/7 on-call, but not support (ie. answering slack questions).

dtuite (changes requested):

src/environment.js
jorgelainfiesta: I ended up not using this, need to remove it. Thanks!

src/pages/index.js
jorgelainfiesta: Gotcha, how about "We offer reliability and 24/7 on-call engineers." Do we offer any concrete uptime?

dtuite:
It's still worth fixing the tests IMO. This is a critical business path and it's good to have it tested in some way.

src/components/landing/CenteredHero/HeroForm.js
dtuite: It's not duplicated.

dtuite (approved):

PR #801, opened 2022-07-26
Add kubecon page placeholder

jorgelainfiesta:
## Motivation

We'll have links from KubeCon pointing to a dedicated page that will be centered on what Roadie can do for Kubernetes users. 

## Approach

We need to commit to a link by July 29th, so I'm adding a route that just redirects to the index while we have the actual page ready.


dtuite (approved):
Very smart!

dtuite (approved):

PR #822, opened 2022-08-10
Add more info to Okta plugin setup page

sblausten:


content/docs/integrations/okta/index.md
```diff
@@ -33,6 +35,10 @@ Set the Okta Organization URL in the Okta configuration page of your Roadie Back
 https://<tenant-name>.roadie.so/administration/settings/okta
 ```
 
+## Data
+
+This plugin uses the email from Users and the name and description from Groups. It will ingest all groups and all users in your Okta organisation.
```
punkle, line 40: It occurs to me that we may want to use other information at some point, e.g. the user name, the name of the user, the profile pic. Im wondering if we should be stating that here. Do you think it might limit our choices in the future?

Thoughts?

sblausten: Can just update it later but maybe we don't need to be explicit anyway. Was just because the Netlify guy asked.

sblausten: I removed the detail

punkle (approved):

PR #838, opened 2022-08-27
Add Rootly Integration

kwent:
## Add Rootly Integration

Cf. https://rootly.com/changelog/2022-06-23-backstage-integration

Source: https://github.com/rootlyhq/backstage-plugin


dtuite:
Thanks for this @kwent. The content is great, but the location is slightly wrong. I've moved your content into our plugin directory, rather than our documentation. You can see the result in #840 .

I'll close this since it is now super-seeded.

kwent:
Thank you 🙏🏻

PR #847, opened 2022-09-20
Add pull request NB

sblausten:
Adds note to github:pull-request doc to make sure they use a new branch.

### DIFF
--- a/content/docs/details/scaffolder/index.md
+++ b/content/docs/details/scaffolder/index.md
@@ -671,6 +671,8 @@ steps:
       targetPath: ./subdirectory
 ```
 
+NB: The branch you use for the pull request must be a new branch for the repo. 
+
 #### Outputs
 
 The `publish:github:pull-request` action produces two outputs.


content/docs/details/scaffolder/index.md
```diff
@@ -671,6 +671,8 @@ steps:
       targetPath: ./subdirectory
 ```
 
+NB: The branch you use for the pull request must be a new branch for the repo.
```
punkle, line 674: I wonder could we use an example that adds random chars to the end of the branch name.

punkle (approved):

content/docs/details/scaffolder/index.md
sblausten: I think as long as they use a roadie / backstage specific branch name the chance of conflicts would be pretty slim

kissmikijr: The branch name collision is going to be a problem when they are experimenting and writing the templates themselves. A straightforward way to test it is to just run the action and see the content in github. During this interaction process they'll need to remember to change the branch name, or increment a counter or something on it.

PR #848, opened 2022-09-20
Only demos as CTA

dtuite:
Switch all CTAs on the site over to demos instead of free trials.

Preview link: https://deploy-preview-848--roadie.netlify.app/


src/components/FormSubmissionModal/FormSubmissionModal.js
```diff
@@ -87,8 +87,8 @@ const FormSubmissionModal = ({
     // Doesn't make sense to offer to let people sign up to the newsletter immediately after
     // they have just signed up to the newsletter.
     followOnContent = <TwitterInner siteMetadata={siteMetadata} />;
-  } else if (followOn === 'GET_DEMO_SURVEY') {
-    followOnContent = <GetDemoSurveyInner referredEmail={email} />;
```
dtuite, line 91: This was named incorrectly. It hasn't been a "survey" for a long time.

src/components/landing/CenteredHero/CenteredHero.js
```diff
@@ -1,52 +1,56 @@
...
-                  {whiteLogos.map((logo) => (
-                    <img
-                      className="m-3 h-8 opacity-80"
-                      key={`hero-logo-${logo.alt}`}
-                      src={logo.src.white}
-                      alt={logo.alt}
-                    />
-                  ))}
-                </div>
+    <div className="relative landing-hero-background overflow-x-hidden">
```
dtuite, line 11: The change in indentation confuses the diff. The only real change is centering the text and moving the CTA to a request demo button.

jorgelainfiesta:
I think the centered text on the second line is hard to read, we'd need to reduce the width of the subtext. Also, I think we could keep the same form on the homepage and just change the button text. Then update what the email they get sent is in Intercom

oblongcoffee (approved):

oblongcoffee:
@dtuite @jorgelainfiesta I've just reaslied that the zapier flow is a bit outdated for demos. It isn't unified like the trial request zapier. It might be worth reviewing it to make sure it is sustainable before setting this live?

PR #871, opened 2022-10-27
Update steps for configuring sonarqube

punkle:
The sonarqube plugin was changed to use a backend instead of a proxy. As such we needed to change how the sonarqube plugin is configued.


content/docs/integrations/sonarqube/index.md
```diff
@@ -26,28 +26,10 @@ This page describes how to create and set up the API token.
...
  * [Sonarcloud](https://sonarcloud.io/account/security) for your sonarcloud plugin
- * [SonarQube](https://docs.sonarqube.org/latest/user-guide/user-token/) for your sonarqube plugin
 
 ### Step 2: Store the credentials and SonarQube URL in Roadie
-Visit `https://<tenant-name>.roadie.so/administration/settings/secrets` and enter the token value from above into `SONARQUBE_TOKEN`.
+Visit `https://<tenant-name>.roadie.so/administration/settings/secrets` and enter the token value from above into `SONARQUBE_API_TOKEN`.
 
-![Token in Roadie](./secret.png)
-
-### Optional Step 3: Set SonarQube proxy in the settings
```
sblausten, line 36: Would we not still need this to connect to an on-prem instance?

punkle: I checked through all of our customers, none are using it, so I decided that I wouldn't undertake supporting this, at this point. It is not a massive amount of work, but would require config pages and testing etc.

I figured if a customer comes along later looking for it, we can add it in.

kissmikijr (approved):

PR #884, opened 2022-11-16
[sc-11760] - add openapi resolver docs

kissmikijr:


kissmikijr:
@dtuite I reopened this as these docs will be needed when we release the v1.8.2 version

sblausten (approved):

PR #909, opened 2022-12-07
refer to steps using dict key lookup

punkle:
We have seen confusion where a step has an id that contains a special character. By documenting this syntax, we hope that users wont have to come across this issue.

### DIFF
--- a/content/backstage/plugins/scaffolder-http-requests.md
+++ b/content/backstage/plugins/scaffolder-http-requests.md
@@ -71,7 +71,7 @@ gettingStarted:
             action: http:backstage:request
             input:
               method: 'POST'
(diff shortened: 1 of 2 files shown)


Xantier (approved):
Can you also write that `-` is not an allowed character in the step name somewhere

punkle:
@Xantier it is an allowed character for step ids. However nunjucks is using it as a special character. I can see examples all over the place using "-" and I doubt we will be able to stop people doing it.. IMO. [(e.g see fetch-base here)](https://backstage.io/docs/features/software-templates/writing-templates)

I agree with the premise of you suggestion though. I think maybe backstage scaffolder could be more strict in its validation of templates before loading them and/or starting them up..

PR #928, opened 2022-12-21
Add docs for broker

Xantier:
* Write broker docs
* Add broker config files to SonarQube and Kubernetesdocs


punkle (approved):

content/docs/docs-nav.yaml
```diff
@@ -53,6 +53,7 @@ nav:
   - Integrations:
       - AWS S3: '/docs/integrations/aws-s3/'
       - Azure DevOps: '/docs/integrations/azure-devops-provider/'
+      - Broker: '/docs/integrations/broker/'
```
punkle, line 56: Feels like this is more like a meads to an integration rather than an integration itself. I wonder is this the best place for this.

I dont have a better suggestion.

Xantier: Yeah, I'm not sure either. It could be a separate configuration section some time in the future maybe?

PR #934, opened 2022-12-26
Upgrade gatsby & associated plugins

dtuite:


dtuite:
Merged without a +1 due to holidays.

PR #936, opened 2022-12-27
Use Roadie <Link /> component

dtuite:
Otherwise external links are not tracked with events and don't open in new tabs.


dtuite:
Merged without review since a link on the homepage is broken.

PR #937, opened 2022-12-27
Warn gmail users on forms

dtuite:
Gmail users never convert so let's try to get them to use their work emails instead.

- [x] Test the free trial flow to ensure it's still working.
- [x] Test the demo flow to ensure it's still working.

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,3 @@
+sonar.sources = src/
+sonar.tests = cypress/
+sonar.cpd.exclusions = cypress/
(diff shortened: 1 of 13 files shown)


sblausten (commented):
Nice. We have a list of common personal emails in the tenant-manager we could use here also to broaden the net. - 

```
[
    'gmail',
    'hotmail',
    'outlook',
    'live',
    'msn',
    'yahoo',
    'aol',
    'gmx',
  ]
  ```

sblausten (approved):

jorgelainfiesta:
Found a very niche use-case where the regex can give a false positive in the zapier integration, but I think it's safe to ignore it: if you have `gmail`, for example, as part of the email (`jorge+test-gmail+trial1@xxxx.xyz`), it'll treat it as a Gmail account. This is unrelated to this PR as it works correctly in the UI with the warning messages. Leaving it here to document it somehow.

jorgelainfiesta (approved):
Both flows work as expected

PR #948, opened 2023-01-13
Product pages

jorgelainfiesta:
## Motivation

Roadie's website needs to talk about more about the product.

## Approach

I've prepared two pages to start explaining the main features of our product: the Catalog and the Scaffolder.


dtuite (approved):

dtuite:
Only nitpick I have is that the navigation to the new product pages is not available on mobile.

PR #949, opened 2023-01-16
Applies outlined text SVGs to screenshots

jorgelainfiesta:
## Motivation

Illustrations featuring text benefit from using an SVG format, so the text will stay clear if the user were to zoom in to read the detail. However, all text should be outlined (turned into paths) instead of being actual text to prevent rendering inconsistencies. 

This is how the current illustration looks in Safari:
Screenshot 2023-01-16 at 12 05 59

## Approach

I outlined the text from every screenshot and replaced the existing SVGs. If an image featured a 'screenshot', I replaced the PNG for an SVG.


Xantier (approved):
Looks grand in Firefox

PR #986, opened 2023-03-22
Plugins nav

jorgelainfiesta:
## Motivation

The marketplace is one of our most significant sources of traffic from SEs, but the individual pages don't have any links to other pages or a clear CTA.

## Approach

In this first iteration, I'm adding navigation to the plugin page and improving the design of the CTA, and other small details. 

## Screenshots
 
Screenshot 2023-03-22 at 17 05 51
Screenshot 2023-03-22 at 17 06 01

## Future work

I want to add related content to the most significant entries, for example the ArgoCD one

### DIFF
--- a/content/backstage/plugins/argo-cd.md
+++ b/content/backstage/plugins/argo-cd.md
@@ -73,7 +73,7 @@ gettingStarted:
       ARGOCD_AUTH_TOKEN='argocd.token=<token>'
 ---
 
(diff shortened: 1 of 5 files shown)


jorgelainfiesta:
Yeah good point; perhaps I can make the link open on a new tab for now. The reason why I want the tabs concept is to communicate that these are alternative methods of installation. Ideally, I'd bring the docs into an actual tab but that would require significant effort.

PR #1034, opened 2023-06-02
Add GA4

dtuite:


dtuite:
https://app.shortcut.com/larder/story/15333/enable-ga4-on-marketing-site

Xantier (approved):

src/googleAnalytics.js
```diff
@@ -11,6 +11,12 @@ const trackGoogleAnalyticsEvent = (eventParams) => {
   }
 
   trackCustomEvent(eventParams);
+
+  if (!isUndefined(window.gtag)) {
+    window.gtag('event', 'pageview', {
+      'dimension1': `'${currentlyExecutingGitBranch()}'`,
```
Xantier, line 17: Worth naming this to 'git_branch' or similar?

dtuite: I suspect it can be deleted. I can't remember why that's there. I think I was debugging in prod. For the sake of this PR, I just want to cut across with feature parity. The deletion should be a separate step.

PR #1043, opened 2023-06-13
Add details about k8s annotations

sblausten:
https://app.shortcut.com/larder/story/15368/bailliegifford-issues-with-k8s-frontend-displaying-pods

### DIFF
--- a/content/docs/integrations/kubernetes/index.md
+++ b/content/docs/integrations/kubernetes/index.md
@@ -22,7 +22,58 @@ Roadie supports the following connection setups:
 
 All three of these can use a Broker service that runs in your infrastructure and means that you do not have to expose your cluster via an AWS cross account Role, Google OAUTH App token or Service Account Token.
 
-# Via Broker
+# Kubernetes Annotations
+
+The k8s plugin uses two alternate ways of discovering resources in addition to a namespace annotation. You can find the Backstage docs on this [here](https://backstage.io/docs/features/kubernetes/configuration/#surfacing-your-kubernetes-components-as-part-of-an-entity).
+
+## 1. backstage.io/kubernetes-id
+You can add a label with the key `backstage.io/kubernetes-id` to all of your k8s resources relating to an entity and then using the same key as an annotation on your entity.
+
+Additionally, you can use a namespace label `backstage.io/kubernetes-namespace` but this does not relate to the actual namespace the resource is deployed in. It is just another label.  
+
+NB: At a minimum for the frontend plugin to show useful data, a deployment and some related resources need the same matching labels.
+
+```yaml
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  ...
+  annotations:
+    backstage.io/kubernetes-id: something
+    backstage.io/kubernetes-namespace: roadie
+...
+```
+
+```yaml
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  ...
+  labels:
+    backstage.io/kubernetes-id: backstage-backend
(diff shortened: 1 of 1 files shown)


content/docs/integrations/kubernetes/index.md
```diff
@@ -22,7 +22,58 @@ Roadie supports the following connection setups:
...
+
+```yaml
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  ...
+  annotations:
+    backstage.io/kubernetes-id: something
+    backstage.io/kubernetes-namespace: roadie
```
dtuite, line 44: Reading to this point, I still don't understand the value of the `backstage.io/kubernetes-namespace` label. Is there something else we can say to explain it? If I remove this, what part of the plugin stops working?

sblausten: I don't either tbh. Its just an additional layer I guess, maybe if you want to keep the id or labels generic then you may need a namespace label to specify exactly what you want to show up and avoid id clashes.

sblausten: I clarified this - I was actually wrong initially on the usage

content/docs/integrations/kubernetes/index.md
```diff
@@ -22,7 +22,60 @@ Roadie supports the following connection setups:
 
 All three of these can use a Broker service that runs in your infrastructure and means that you do not have to expose your cluster via an AWS cross account Role, Google OAUTH App token or Service Account Token.
 
-# Via Broker
+# Kubernetes Annotations
+
+The k8s plugin uses two alternate ways of discovering resources in addition to a namespace annotation. You can find the Backstage docs on this [here](https://backstage.io/docs/features/kubernetes/configuration/#surfacing-your-kubernetes-components-as-part-of-an-entity).
+
+## 1. backstage.io/kubernetes-id
+You can add a label with the key `backstage.io/kubernetes-id` to all of your k8s resources relating to an entity and then using the same key as an annotation on your entity.
```
iain-b, line 30: I'm assuming the value is arbitrary?

content/docs/integrations/kubernetes/index.md
```diff
@@ -22,7 +22,60 @@ Roadie supports the following connection setups:
...
+
+Additionally, you can lookup resources by namespace using the `backstage.io/kubernetes-namespace` annotation.
```
iain-b, line 32: I'm assuming the default is "default"

content/docs/integrations/kubernetes/index.md
```diff
@@ -22,7 +22,60 @@ Roadie supports the following connection setups:
...
+kind: Component
+metadata:
+  ...
+  annotations:
+    backstage.io/kubernetes-id: something
+    backstage.io/kubernetes-namespace: roadie
+...
+```
+
+```yaml
```
iain-b, line 48: perhaps a comment just to highlight that this is a k8s resource and the above is a Backstage entity ? They have the same structure so I can see how users might get confused.

iain-b (approved):
Lgtm with some nit picks

content/docs/integrations/kubernetes/index.md
sblausten: Yes in their example they use the entity name but it just searches as if it were any other label.

content/docs/integrations/kubernetes/index.md
sblausten: Oh well I just tested that and I think I'm wrong there.

sblausten: No that was just a misconfig

PR #1047, opened 2023-06-14
Add context on GitHub token

dtuite:


sblausten (approved):

content/docs/tech-insights/builtin-data-sources/index.md
```diff
@@ -150,8 +150,9 @@ If you have GitHub App installed, access will be granted based on permissions gr
 Generates fact data about GitHub for the entity
 
 **Annotations used:** `github.com/project-slug`
+
 #### Authentication
-Authentication is done via `GITHUB TOKEN` [Read more](../../integrations/github-token/). 
+Authentication is done via `GITHUB TOKEN`. We use a token, rather than the Roadie GitHub App, because we don't know in advance which GitHub APIs you will need permissions to hit with the data sources you create. When you create and use your own token, you can pick and choose which permissions to give it, depending on your use cases. [Learn how to set a token](../../integrations/github-token/).
```
Xantier, line 155: If this is true (I don't know if it is, it shouldn't be), it is a bad UX and should be fixed. WE write the builtin GitHub fact retriever, so we definitely know which APIs we are hitting with it. 

If we want to keep plumping up the GitHub app permissions list to achieve this is a different story.

cc @Irma12 @dtuite

PR #1105, opened 2023-08-24
Adds a CTA to the Scaffolder doc

jorgelainfiesta:
## Motivation

The Writing Scaffolder Templates page drives traffic from Google Search to the website, but there's no CTA on it.

## Approach

I added an HTML and CSS prompt to the docs. The link includes UTM parameters so we can hopefully track if this page converts leads. 

## Preview
Screenshot 2023-08-25 at 11 27 23

https://deploy-preview-1105--roadie.netlify.app/docs/scaffolder/writing-templates/

### DIFF
--- a/content/docs/scaffolder/writing-templates/index.md
+++ b/content/docs/scaffolder/writing-templates/index.md
@@ -15,10 +15,7 @@ A Scaffolder template is then run on demand by the users of Backstage to execute
 You can find a step by step guide to adding templates in Roadie [here](/docs/getting-started/scaffolding-components/)
 
 <div class="docs-cta">
-  <h4 class="docs-cta__title">See the Scaffolder in action</h4>
-  <p class="docs-cta__paragraph">
-    Discover more self-service possibilities by the hand of a Backstage expert. 
-  </p>
+  <h4 class="docs-cta__title">See the Backstage Scaffolder in action</h4>
   <a href="/request-demo/?utm_source=roadie-docs&utm_campaign=scaffolder-docs" target="_blank" class="docs-cta__button">
     Request a Roadie demo
   </a>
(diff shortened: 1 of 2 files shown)


content/docs/scaffolder/writing-templates/index.md
```diff
@@ -14,6 +14,16 @@ A Scaffolder template is then run on demand by the users of Backstage to execute
 
 You can find a step by step guide to adding templates in Roadie [here](/docs/getting-started/scaffolding-components/)
 
+<div class="docs-cta">
+  <h4 class="docs-cta__title">Want to see the Scaffolder in action?</h4>
```
dtuite, line 18: I'd use more direct language here:

```suggestion
  <h4 class="docs-cta__title">See the Scaffolder in action</h4>
```

content/docs/scaffolder/writing-templates/index.md
```diff
@@ -14,6 +14,16 @@ A Scaffolder template is then run on demand by the users of Backstage to execute
...
+  <p class="docs-cta__paragraph">
+    Discover more self-service possibilities by the hand of a Backstage expert.
```
dtuite, line 20: The English is slightly off here. Something like "in the hands of" would read a little more native.

You might be able to do without this sentence entirely though. Perhaps "See the Scaffolder in action" is enough?

jorgelainfiesta: You're right, I was trying to force the "Backstage expert" part. I do want to have a little paragraph here so they don't think they'll be going to some kind of playground

jorgelainfiesta: I ended up removing it because the page they get to already has this copy so let's bet for shorter as you suggested.

dtuite (approved):

PR #1159, opened 2023-11-02
Add Roadie API Docs

kissmikijr:


Xantier:
Nice, that came out smoothly. Are we able to hide/remove the 'Authorize' btn somehow? Maybe adding global CSS to just display:none on that specific item

kissmikijr:
Yes, with a targeted CSS we can hide it.

PR #1176, opened 2023-11-27
Use v7 of sub-processors

dtuite:
https://app.shortcut.com/larder/story/17739/update-sub-processors-on-roadie-io


oblongcoffee (changes requested):
I would remove Intercom too

oblongcoffee (approved):

PR #1177, opened 2023-11-28
add docs for accessing the user entity

punkle:


content/docs/scaffolder/writing-templates/index.md
```diff
@@ -473,18 +489,24 @@ If the step id does not contain a special character you can also refer to it usi
...
+#### Accessing the logged in user
+
+You can refer to the user entity reference for the logged in user using the following syntax:
+
+```yaml
+${{ user.ref }}
+```
 
-You can refer to the value of a parameter using the following syntax:
+If this entity reference [exists in the backstage catalog](/docs/getting-started/teams/), you can also make use of the details contained within the users entity by using the following:
```
Xantier, line 500:
```suggestion
If this entity reference [exists in the Backstage Catalog](/docs/getting-started/teams/), you can also make use of the details contained within the users entity by using the following:
```

Xantier (approved):

dtuite (approved):
Super cool. Didn't know you could do this

PR #1194, opened 2024-01-04
Remove snyk annotations

dtuite:


catalog-info.yaml
```diff
@@ -17,8 +17,6 @@ metadata:
   annotations:
     github.com/project-slug: RoadieHQ/marketing-site
     backstage.io/techdocs-ref: dir:.
```
punkle, line 19: I think you need project ids no?

punkle (approved):

catalog-info.yaml
dtuite: I thought it would fall back to the project-slug

punkle: oh yeah. I forgot about that one.

PR #1202, opened 2024-01-16
Add documentation for roadie api

punkle:


sblausten (approved):
Nice one. Have we released the self service token generation yet though? Maybe we need to wait to merge this until thats done if not

samnixon87 (approved):
Nice. Same comment as @sblausten - we should time this with the launch of self-serve token generation.

Xantier:
Was this merge premature?

kissmikijr:
I think so

punkle:
@Xantier @kissmikijr why was it premature? My understanding was that we were including the docs in the launch silently step: https://roadiehq.slack.com/archives/CUFAGVDDG/p1705410608282399

PR #1205, opened 2024-01-19
Add back in the servers list to the api docs

punkle:
Before

image

After:

image


Ian-Link-Roadie (approved):
We should discuss versioning soon. As the other API's are versioned to v1 or v1.0.0 already.

PR #1251, opened 2024-03-26
Add z-index to movile nav

dtuite:
## Before

The sitewide nav is open but it's hidden behind the title of an article.

Screenshot 2024-03-26 at 20 27 20

## After

Screenshot 2024-03-26 at 20 27 28


dtuite:
https://app.shortcut.com/larder/story/19795

Xantier (approved):

PR #1258, opened 2024-04-03
TLDR section prototype

Ian-Link-Roadie:
Trying a few formatting options across the Jenkins, ArgoCD, and PagerDuty tutorials.

Preview link:

1. https://deploy-preview-1258--roadie.netlify.app/docs/integrations/pagerduty/


Ian-Link-Roadie:
https://app.shortcut.com/larder/story/19926/create-plug-in-tl-dr-section

dtuite:
Feedback

1. The layout of the table looks good
2. The "Expectations" should come after the Introduction. We have to tell the user what the benefit will be before we tell them about the work they need to do.
3. I'm not sure Expectations is the right title for this section. Something like "Summary" or "Details" makes more sense to me. 
4. "Environments" should be "Supported environments"

Ian-Link-Roadie:
> Feedback
> 
> 1. The layout of the table looks good
> 2. The "Expectations" should come after the Introduction. We have to tell the user what the benefit will be before we tell them about the work they need to do.
> 3. I'm not sure Expectations is the right title for this section. Something like "Summary" or "Details" makes more sense to me.
> 4. "Environments" should be "Supported environments"

What do you think about "At a Glance" for the section name? I'll move it below the Intro and update the Env row name.

PR #1260, opened 2024-04-11
mention section index fix

punkle:


content/docs/details/techdocs/index.md
```diff
@@ -77,6 +77,15 @@ page structure using the `nav` object in your `mkdocs.yaml`. Both approaches are
...
 
+By default MkDocs will create a navigation menu where a menu item that has child pages will not link to the index page for that parent item. To fix this we have added a MkDocs plugin called `section-index`. To enable this plugin you can add it to the plugins list in the `mkdocs.yaml` file.
+
+
+```yaml
+plugins:
+  - section-index
+  - ...
+```
+
```
samnixon87, line 88: I could be wrong but afaict from here https://github.com/oprypin/mkdocs-section-index/blob/master/mkdocs_section_index/__init__.py you can pass some stuff into sections when you create them. we should tell people about that if it's possible

punkle: This explanation is already linked from the docs here:

example

PR #1359, opened 2024-08-20
Adding docs for new sidebar settings

roadgork:


content/docs/details/updating-the-ui/add-page.jpeg
Xantier, line 1: Can you change this to have a better title

content/docs/details/updating-the-ui/add-custom-page-for-sidebar.jpeg
Xantier, line 1: Can you modify the image itself. It's displaying our detrimental approach of splitting up words randomly on capital letters. The actual title value we would want to show on this input field is `GraphiQL` and the path would be something like `graphiql-page`

Xantier (approved):

PR #1369, opened 2024-08-30
Add referrer to trial and demo requests

dtuite:


dtuite:
https://app.shortcut.com/larder/story/22581/add-referrer-to-demo-and-trial-requests

src/pages/whitepapers/managed-vs-self-hosted-backstage.tsx
```diff
@@ -23,7 +23,7 @@ const RequestTrial = ({ data, location }) => {
       />
 
       <div className="min-h-screen bg-white">
-        <SitewideHeader ctaText="Request a Demo" ctaTo="/request-demo/" />
+        <SitewideHeader ctaText="Request a Demo" />
```
dtuite, line 26: This is the default in the `<SitewideHeader />`

package.json
```diff
@@ -111,7 +115,7 @@
     "test": "cypress run",
     "lint": "eslint src",
     "preview": "yarn run build && netlify deploy --dir public",
-    "test:unit": "mocha --require @babel/register src/**/*.test.js",
+    "test:unit": "jest",
```
dtuite, line 118: I switched all unit tests to jest instead of mocha because I needed to use full DOM rendering (rather than shallow rendering) in the tests of the `<Link />` component. Jest has better integrations with JSDOM for doing this. There wasn't many unit tests in the site so it didn't take long.

src/components/Link/Link.js
```diff
@@ -48,6 +73,8 @@ const Link = ({
   forceOpenInSameTab = false,
   ...rest
 }) => {
+  const location = useLocation();
```
dtuite, line 76: Now that this component uses a hook which relies on looking up the location on the window object, we need to provide a window like environment for testing. Hence the switch to jest and JSDOM for the tests of this component. They are probably better tests now anyway, since they run in an environment more like userland.

PR #1385, opened 2024-09-16
Serve trailing slashes in sitemap

dtuite:
roadie.io uses paths with a trailing slash by default. Every non-trailing slash path is set up to redirect to a trailing slash path.

Our sitemap, however was set up to serve paths without a trailing slash. This means that we are serving 301 redirects to crawlers like Google on every path of the site.

There's no way to fix this while using `gatsby-plugin-advanced-sitemap` ([issue here](https://github.com/TryGhost/gatsby-plugin-advanced-sitemap/issues/223)) so I have switched us to `gatsby-plugin-sitemap` ([docs here](https://www.gatsbyjs.com/plugins/gatsby-plugin-sitemap/)) which is the default for gatsby.

## Before

Screenshot 2024-09-16 at 11 49 39

## After

Screenshot 2024-09-16 at 11 51 05


dtuite:
https://app.shortcut.com/larder/story/22822/fix-sitemap

Xantier (approved):

PR #1389, opened 2024-09-19
Add new sitemap for images to resolve indexing issues with webp images

samnixon87:
NB: May increase build time

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -57,28 +57,6 @@ const getContentfulOptions = () => {
   };
 };
 
-// These paths are blocked from search engine indexing and search engine access.
-const DISLALLOW_LIST = [
-  '/purchase/',
-  '/purchase/success/',
-  '/tailwind/404/',
-  '/installation-pending/',
-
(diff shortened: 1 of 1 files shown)


gatsby-config.js
```diff
@@ -209,7 +209,34 @@ module.exports = {
     {
       resolve: 'gatsby-plugin-sitemap',
       options: {
+        output: `/sitemap-index`,
         excludes: DISLALLOW_LIST,
```
Xantier, line 213: just saw this. Can you rename

samnixon87: afaik it should match the other sitemaps, which are currently only on sitemap-index. sitemap-index should collect all sitemaps though, so I'll change and see what happens

Xantier: Meant the type here. DIESEL-ALLOW

dtuite:
Test this against the roadie.io/sitemap-0.xml sitemap please. We need to be sure we don't actually deindex a bunch of pages or something like that.

Xantier (approved):

PR #1401, opened 2024-10-03
Scaffolder actions

sblausten:
https://app.shortcut.com/larder/story/22851/scaffolder-actions-comprehensive-resource


sblausten:
Need to add whether each one is available in Roadie

content/docs/scaffolder/scaffolder-actions-directory/index.md
```diff
@@ -5,18 +5,166 @@ description: Directory of available scaffolder actions
...
+| Yeoman               | [run:yeoman](#runyeoman)                  | Action for running a Yeoman generator in Backstage scaffolder.                                                                                                            | [Yeoman Run](https://github.com/backstage/backstage/tree/master/plugins/scaffolder-backend-module-yeoman)                                                                                                                                                                             | namespace*, args, options                                                                               |
+
+## Download content
```
Xantier, line 153: This title doesn't seem like it fits here

content/docs/scaffolder/scaffolder-actions-directory/index.md
```diff
@@ -5,18 +5,166 @@ description: Directory of available scaffolder actions
 ---
 
 
-## Directory of available scaffolder actions
-
-In this page you can find a directory of some available Scaffolder actions and their examples. You can find a full list of available actions to your Roadie instance by visiting the following page from within Roadie:
-
-`https://<tenant-name>.roadie.so/templates/actions`
+# Directory of available scaffolder actions
```
punkle, line 8: There is a difference between the actions available in Roadie and the actions available for backstage. For the plugins we have two difernent pages for this.. the open source plugins directory for backstage is under resources off the main page.

By documenting all of the actions here I think implies that these actions are available on roadie.

sblausten: Yea I'll change "available". I'm not sure splitting them up is very helpful though from a SEO standpoint. I prefer to mark which actions are available in Roadie as mentioned in my comment. I think we will be revising this content into more digestible formats anyway.

content/docs/scaffolder/scaffolder-actions-directory/index.md
sblausten: What category would you call these then?

Xantier: It's not the category. There is a full title missing. Now it's abruptly changing into a different list after the table.

content/docs/scaffolder/scaffolder-actions-directory/index.md
punkle: Perfect 👌 , so long as its clear. btw.. the roadie supported actions directory is already here https://roadie.io/docs/scaffolder/scaffolder-actions-directory/

So I guess you are creating duplication by creating this page?

content/docs/scaffolder/scaffolder-actions-directory/index.md
sblausten: Is that better now?

content/docs/scaffolder/scaffolder-actions-directory/index.md
sblausten: No you can see in the changelog that this is an update to the original page

content/docs/scaffolder/scaffolder-actions-directory/index.md
Xantier: yup, looks good :+1:

PR #1402, opened 2024-10-08
Add docs on how to enable finding ECS Services based on tags.

Xantier:


content/docs/integrations/aws-plugins/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Configuring AWS Plugins
-publishedDate: '2022-03-29T21:00:00.0Z'
+publishedDate: '2024-10-08T10:51:37.0Z'
```
punkle, line 3: We are we updating this? Should we use an updated date or something like that?

PR #1410, opened 2024-10-18
Make Jira token generation docs simpler

sblausten:
https://app.shortcut.com/larder/story/23336/actblue-jira-plugin-setup-problems

### DIFF
--- a/content/docs/integrations/jira/index.md
+++ b/content/docs/integrations/jira/index.md
@@ -52,7 +52,7 @@ amlyYS1tYWlsQGV4YW1wbGUuY29tOmhUQmdxVmNyY3hSWXBUNVRDelRBOUMwRg==
 ```
 Via the following command:
 ```bash
-read -s tmp; echo -n "jira-mail@example.com:$tmp" | base64 ; unset tmp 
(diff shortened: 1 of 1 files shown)


Xantier (approved):
We have these methods listed in AWS resources side:
```
* Terraform: `base64encode("roadie-123456789012")`
* Pulumi (JS/TS): `Buffer.from('roadie-123456789012').toString('base64')`
* Bash: `echo -n 'roadie-123456789012' | base64`
* Browser Devtools: `btoa("roadie-123456789012")`
```

Maybe worth adding here also?

sblausten:
> We have these methods listed in AWS resources side:
> 
> ```
> * Terraform: `base64encode("roadie-123456789012")`
> * Pulumi (JS/TS): `Buffer.from('roadie-123456789012').toString('base64')`
> * Bash: `echo -n 'roadie-123456789012' | base64`
> * Browser Devtools: `btoa("roadie-123456789012")`
> ```
> 
> Maybe worth adding here also?

I think in this case they would unlikely to be using Terraform or Pulumi for a manual step like this actually

PR #1421, opened 2024-10-30
fix dead backlink

punkle:


netlify.toml
```diff
@@ -81,6 +81,11 @@
   to = "/docs/details/allowlisting-roadie-traffic/"
   force = true
 
+[[redirects]]
+  from = "/docs/getting-started/openapi-specs"
+  to = "/docs/api"
```
dtuite, line 86: We should be redirecting to the trailing slash version of this. Otherwise it will just redirect again. eg https://roadie.io/docs/api

```suggestion
  to = "/docs/api/"
```

dtuite (changes requested):

punkle:
@dtuite done

sblausten (approved):

netlify.toml
```diff
@@ -81,6 +81,11 @@
   to = "/docs/details/allowlisting-roadie-traffic/"
   force = true
 
+[[redirects]]
+  from = "/docs/getting-started/openapi-specs"
+  to = "/docs/api/authorization/"
```
dtuite, line 86: Actually, this doesn't make any sense. 

The broken page is about how to add API specs to the catalog. The page you're replacing it with is about authenticating with the Roadie APIs.

These are completely unrelated. You need to find a different page to redirect to.

punkle: Will I go back to what i had?

dtuite (approved):

PR #1477, opened 2025-03-04
Update Wiz integration status to certified

jianreis:
Changing the Wiz docs to reflect it's certified as part of the push to publicize this.


content/docs/integrations/wiz/index.md
```diff
@@ -8,9 +8,9 @@ logoImage: '../../../assets/logos/wiz/wiz-logo.png'
 integrationType: OSS plugin
 ---
 
-## In Development
+## Certified Plugin
 
-This plugin is currently in development and awaiting certification from Wiz.
+This plugin is [certified by Wiz](https://www.wiz.io/integrations?q=roadie).
```
samnixon87, line 13: We should add the `Wiz Certified` logo here from their partnership asset pack.

dtuite:
I'm going to put a +1 on this to unblock it. The logo would be better, but what we have is better than what's there currently.

dtuite (approved):

PR #1490, opened 2025-04-01
Add Roadie Local to pricing page

samnixon87:
New Pricing widget, flips between SaaS and On-prem.

image

Also added Roadie Local to the features table.

All copy / prices up for debate of course.

image


Xantier (approved):

dtuite (changes requested):
Blocking until after KubeCon

PR #1509, opened 2025-06-12
Newsletter recaptcha

dtuite:
We're getting spam in the newsletter sign up. I need to turn on recaptchas, but I don't want to turn it on in the sales funnel because there's no spam there so it's just extra friction. This splits the recaptcha toggle in two so I can enable it on the newsletter only.


dtuite:
https://app.shortcut.com/larder/story/27160

Xantier (approved):
Looks fine. AI thinks there are few nits

PR #1511, opened 2025-06-19
Gitlab Oauth docs

iain-b:
Add docs describing how to create a GitLab Oauth app and configure in Roadie.


Xantier (approved):

Xantier:
I think we use terminology `Roadie` instead of `Roadie Backstage` elsewhere, maybe worth a double check.

PR #1523, opened 2025-06-30
Test Roadie Local instructions

Ian-Link-Roadie:
Story details: https://app.shortcut.com/larder/story/27499

### DIFF
--- a/content/docs/details/roadie-local/index.md
+++ b/content/docs/details/roadie-local/index.md
@@ -52,17 +52,17 @@ To test the installation, ensure this command displays help text.
 1. Initialize the environment and enter your license key
 
 ```bash
-./roadie-local start
+./roadie-local init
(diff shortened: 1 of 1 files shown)


content/docs/details/roadie-local/index.md
```diff
@@ -52,17 +52,17 @@ To test the installation, ensure this command displays help text.
 1. Initialize the environment and enter your license key
 
 ```bash
-./roadie-local start
+./roadie-local init
```
samnixon87, line 55: I'm not sure this shouldn't be start. What I'd suggest is that we list the config yaml as a pre-req, then potentially state the error

Xantier: The documentation is clearly wrong at the moment. I will merge this in without exploding the scope.

PR #1527, opened 2025-07-05
Update USD conversion rate

dtuite:


dtuite:
https://app.shortcut.com/larder/story/27560

PR #1528, opened 2025-07-07
Add docs for authorization API write operations

Xantier:


content/docs/api/authorization/index.md
```diff
@@ -23,3 +23,15 @@ curl \
...
   https://api.roadie.so/api/catalog/entities
 ```
+
+For write operations using PUT, POST, or PATCH requests with a request body, we expect JSON structures. You should modify your calls to include the `Content-Type` header.
+
+```shell
+curl \
+  -X POST \
+  -H 'Accept: application/json' \
+  -H 'content-type: application/json' \
```
punkle, line 33: To make it consistent with the other headers.

```suggestion
  -H 'Content-Type: application/json' \
```

PR #1550, opened 2025-08-01
Add more vscode docs for mcp servers

samnixon87:
VSCode does a weird thing with the readOnlyTrue flag. We should tell people that

### DIFF
--- a/content/docs/details/roadie-mcp/index.md
+++ b/content/docs/details/roadie-mcp/index.md
@@ -97,7 +97,7 @@ Open VS Code and try asking Copilot questions like:
 
 #### Skipping steps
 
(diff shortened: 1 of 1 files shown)


content/docs/details/roadie-mcp/index.md
```diff
@@ -95,6 +95,11 @@ Open VS Code and try asking Copilot questions like:
 - "What APIs are available for user management?"
 - "Who owns the payment-service component?"
 
+#### Skipping steps
+
+- VSCode omits some information-only steps and/or auto-completes various actions our MCP tools request. That is due to a permissive interpretation of the protocols `readOnlyHint: true` flag that it is best practice to use on MCP servers [based on the protocol specification](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations). The flag represents non-destructive tools which only return information and do not alter the MCP clients environment. VSCode interprets `readOnlyHints` as default permissable to execute, whereas most other MCP clients require user consent or a flag to be set in config before they autocomplete.
```
Xantier, line 100:
```suggestion
- VSCode omits some information-only steps and/or auto-completes various actions our MCP tools request. That is due to a permissive interpretation of the protocols `readOnlyHint: true` flag, which is best practice to use on MCP servers [based on the protocol specification](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations). The flag represents non-destructive tools which only return information and do not alter the MCP clients environment. VSCode interprets `readOnlyHints` as default permissable to execute, whereas most other MCP clients require user consent or a flag to be set in config before they autocomplete. 
```

The grammar went off I think

PR #1555, opened 2025-08-11
Remove price selection dropdown

dtuite:


dtuite:
https://app.shortcut.com/larder/story/28100/update-pricing-page

src/components/pricing/TierBulletsSection.js
```diff
@@ -1,14 +1,22 @@
 import React from 'react';
-import { CheckIcon } from '@heroicons/react/solid'
+import { CheckIcon, PlusIcon } from '@heroicons/react/solid';
 
-const TierBullet = ({ feature, hasIcon = true }) => (
+const renderIcon = ({ iconName }) => {
```
dtuite, line 4: Is there a more "React" way to dynamically choose an icon like this?

Xantier: You'd make this `renderIcon` function into a component that takes a prop of the icon name. For typescript you also want to restrict the available icon names to the actually supported strings.

Xantier (approved):

PR #1563, opened 2025-08-28
document the split function

punkle:


content/docs/scaffolder/recipes/split-string/index.md
```diff
@@ -0,0 +1,94 @@
+---
```
punkle, line 1: 100% slop. I proof read it.

roadgork (approved):

PR #1578, opened 2025-09-12
Add NPM data to plugins directory

dtuite:
Example of changed URLs:

- https://deploy-preview-1578--roadie.netlify.app/backstage/plugins/incident-io/
- https://deploy-preview-1578--roadie.netlify.app/backstage/plugins/api-docs/
- https://deploy-preview-1578--roadie.netlify.app/backstage/plugins/dynatrace/

### DIFF
--- a/src/components/backstage/plugins/Sidebar.js
+++ b/src/components/backstage/plugins/Sidebar.js
@@ -26,7 +26,7 @@ const DetailsListItem = ({ label, value, ...props }) => {
   );
 };
 
-const Mainatainer = ({ name, email }) => {
+const Maintainer = ({ name, email }) => {
   return (
     <li title={`Name: ${name}. Email: ${email}`}>
       <GravatarImage email={email} name={name} />
@@ -60,7 +60,7 @@ const MaintainersList = ({ npmData, npmDataLoadingState }) => {
       <>
         <ul className="grid grid-cols-4 gap-3 pb-1">
           {map(npmData.maintainers, ({ name, email }) => (
-            <Mainatainer name={name} email={email} key={email} />
+            <Maintainer name={name} email={email} key={email} />
           ))}
         </ul>
         <p className="italic text-gray-400">
--- a/src/npmPackageData/stripPackageData.test.mjs
+++ b/src/npmPackageData/stripPackageData.test.mjs
@@ -37,6 +37,6 @@ describe('#stripPackageData', () => {
   });
 
   it('should not include the versions object', () => {
-    expect(stripPackageData(testPackage).versions).toBeUndefined;
+    expect(stripPackageData(testPackage).versions).toBeUndefined();
   });
 });


src/components/Byline/Byline.test.js
dtuite, line 1: Changes in this file are not directly related to the linked ticket. I ended up going down a bit of a rabbit hole to get tests to pass. Regardless, the change is worthwhile because the `Enzyme` testing library [is deprecated and doesn't support React 18](https://stackoverflow.com/a/73999396). 

I ended upgrading `jest` which allowed me to use it's support for freezing time. This was the only reason `sinon` was being installed, so that was no longer necessary.

src/templates/Plugin.js
```diff
@@ -82,13 +78,23 @@ const PluginTemplate = ({ data }) => {
...
-          <PluginCTA plugin={plugin} />
+      <div className="mt-4">
+        <Header plugin={plugin} />
+      </div>
+
+      <main className="pb-8 px-4 lg:pb-28">
+        <div className="relative max-w-7xl mx-auto">
+          <div className="grid grid-cols-3 gap-20">
+            <article className="col-span-2">
+              <Body plugin={plugin} siteMetadata={siteMetadata} />
```
dtuite, line 89: Next improvement will probably split all this content into tabs to make the pages easier to read.

src/templates/Plugin.js
```diff
@@ -161,3 +167,24 @@ export const pageQuery = graphql`
...
+    data = await retrievePackageDataByName({
+      packageName: pageContext.npmjsPackage,
+      authStrategy: 'token',
+    });
+
+  } catch (e) {
+    // The page is still useful without the NPM data so just log the error
+    // and return a 200 code and the UI can handle the fact that the NPM
+    // data is missing.
+    console.error(e);
```
dtuite, line 184: I'm assuming this will swallow the error and not rethrow it. Is that true?

package.json
```diff
@@ -56,6 +56,7 @@
     "gatsby-transformer-remark": "^6.14.0",
     "gatsby-transformer-sharp": "^5.14.0",
     "husky": "^7.0.4",
+    "lodash": "^4.17.21",
```
dtuite, line 59: This should always have been here. Aparently we were relying on a transitive dep.

src/components/backstage/plugins/Sidebar.js
```diff
@@ -0,0 +1,171 @@
...
+
+  try {
+    response = await fetch(`${funcUrl}?packageName=${packageName}`);
+  } catch (err) {
+    console.error(err);
+    return data;
+  }
+
+  try {
+    const json = await response.json();
```
punkle, line 51: I think you want to also check the response code is ok before attempting to return the data. Otherwise you might return some error content as a valid result:

```suggestion
    if (!response.ok) {
      return data;
    }
    
    const json = await response.json();
```

dtuite: Fixed in a future commit.

gatsby-config.mjs
```diff
@@ -39,10 +41,14 @@ const skipWebpackAnalyzer = has(process.env, 'GITHUB_ACTIONS') || has(process.en
 const getContentfulHost = () => {
   if (has(process.env, 'GITHUB_ACTIONS')) return 'cdn.contentful.com';
 
-  // This is an environment variable set by the Netlify build process.
-  const netlifySiteName = get(process.env, 'SITE_NAME');
-  if (netlifySiteName === 'roadie-preview') return 'preview.contentful.com';
-  if (netlifySiteName === 'roadie') return 'cdn.contentful.com';
+  // This environment variable is always set when building in the netlify environment. It is
+  // not set by the netlify dev server.
+  if (get(process.env, 'NETLIFY')) {
```
dtuite, line 46: The presence of the Netlify `SITE_NAME` env var was being used to check if we are executing on netlify (aka a production like environment). However, this broke use of the netlify dev server `yarn netlify dev` because that also sets the `SITE_NAME` env var. To make this more explicit, I added a check for the `NETLIFY` env var, whcih is only present when we are executing on Netlify.

gatsby-config.mjs
```diff
@@ -99,7 +105,9 @@ process.env.GATSBY_GIT_BRANCH_NAME = process.env.BRANCH;
 process.env.GATSBY_NETLIFY_SITE_NAME = process.env.SITE_NAME;
 process.env.GATSBY_SITE_RECAPTCHA_KEY = process.env.SITE_RECAPTCHA_KEY;
 
-module.exports = {
+const __dirname = dirname(fileURLToPath(import.meta.url))
```
dtuite, line 108: `__dirname` is not automatically available in ESM

package.json
```diff
@@ -104,19 +103,18 @@
   },
   "scripts": {
     "prebuild": "node bin/fetch_api_docs_and_roadie_local.js",
-    "build": "yarn clean && yarn prebuild && gatsby build",
+    "build": "yarn prebuild && gatsby build",
```
dtuite, line 106: Using clean is now a bad practice: https://www.gatsbyjs.com/docs/how-to/performance/improving-build-performance/#make-sure-youre-not-clearing-the-cache-between-builds

package.json
```diff
@@ -104,19 +103,18 @@
...
     "tech-docs:build": "bash bin/convert_tech_docs.sh",
     "tech-docs:clean": "rm -rf roadie-tech-docs",
     "develop": "gatsby develop",
     "format": "prettier --write \"**/*.{js,jsx,ts,tsx,json,md}\"",
-    "start": "gatsby develop",
-    "start-with-env": "env $(cat .env | xargs) yarn start",
+    "start": "netlify dev",
```
dtuite, line 111: Using the netlify dev server instead of bare gatsby provides a local implementation of netlify functions and the netlify blob store. The site depends on these now so it's good to have them available during development.

package.json
```diff
@@ -104,19 +103,18 @@
...
     "serve": "gatsby serve",
     "clean": "gatsby clean",
     "test": "cypress run",
     "lint": "eslint src",
     "preview": "yarn run build && netlify deploy --dir public",
-    "test:unit": "jest",
+    "test:unit": "node --experimental-vm-modules $(yarn bin jest)",
```
dtuite, line 117: Required to make Jest work with ESM. https://jestjs.io/docs/ecmascript-modules

src/npmPackageData/retrievePackageDataByName.mjs
```diff
@@ -0,0 +1,16 @@
+import getRoadieStore from './getRoadieStore.mjs';
+
+const retrievePackageDataByName = async ({
+  packageName,
+}) => {
+  if (typeof packageName === 'undefined' || packageName === '') {
+    throw new Error(`
+      Argument packageName '${packageName}' must not be undefined or an empty string.
```
dtuite, line 8: Netligy blobs seems to return the whole datastore contents if you use `#get` with `undefined` in the first argument. 🤷

src/theme.mjs
```diff
@@ -10,10 +10,12 @@
...
+export default {
   COLORS_PRIMARY_500: '#f97316',
   COLORS_PRIMARY_600: '#ea580c',
   COLORS_GRAY_500: '#737373',
-  BREAKPOINTS_MD: '768px',
-  BREAKPOINTS_LG: '1024px',
+  BREAKPOINTS_MD: '48rem',
+  BREAKPOINTS_LG: '64rem',
+  BREAKPOINTS_XL: '80rem',
+  BREAKPOINTS_2XL: '96rem',
```
dtuite, line 20: This is how tailwind defines breakpoints now: https://tailwindcss.com/docs/responsive-design#targeting-a-breakpoint-range

.github/workflows/test.yml
```diff
@@ -47,11 +47,13 @@ jobs:
...
           ROADIE_API_TOKEN: ${{ secrets.ROADIE_API_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.RELEASES_GITHUB_TOKEN }}
+          NETLIFY_API_TOKEN: ${{ secrets.NETLIFY_API_TOKEN }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
 
       - name: Cypress run
         uses: cypress-io/github-action@v4
         with:
-          start: yarn serve --port 8001
+          start: yarn serve --port 8888
```
dtuite, line 56: Increases the consistency between github actions and local development.

cypress.config.mjs
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from 'cypress';
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:8888',
+    // Has to be wide enough for some responsive elements which only render on desktop screens
+    // to render.
+    viewportWidth: 1200,
```
dtuite, line 8: I had to make the viewport wide in the cypress environment so that the responsive-ly designed sidebar would appear. It's hidden at smaller viewports.

PR #1641, opened 2025-10-20
Cache npm deps in GitHub actions

dtuite:


.github/workflows/test.yml
```diff
@@ -15,17 +15,17 @@ jobs:
       NODE_ENV: test
 
     steps:
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@v5
+      - uses: actions/setup-node@v5
         with:
           node-version: 22.x
+          cache: 'npm'
```
punkle: Should this be `cache: 'yarn'`

PR #1651, opened 2025-10-25
Preserve query when navigating plugins directory. Add icons

dtuite:
## Before

[Test here](https://roadie.io/backstage/plugins/)

1. Starting here, type something in the "Search" input, click a plugin, then press the browser back button. The search query would be lost.
2. There are no icons on the inputs.
     Screenshot 2025-10-25 at 19 37 41
3. The list has no empty state
    Screenshot 2025-10-25 at 19 38 17

## After

[Test here](https://deploy-preview-1651--roadie.netlify.app/backstage/plugins/)

1. Search queries are preserved when navigating into plugins and back
2. Inputs have icons and the list has an empty state.

    Screenshot 2025-10-25 at 19 39 13

### DIFF
--- a/src/templates/Plugin.js
+++ b/src/templates/Plugin.js
@@ -18,7 +18,7 @@ const recordExitIntentModalHasBeenShown = () => {
   return localStorage.setItem('exitIntentModalHasBeenShown', true);
 };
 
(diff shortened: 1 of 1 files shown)


dtuite:
https://app.shortcut.com/larder/story/29749

PR #1653, opened 2025-10-26
Prevent `it.only` being used in CI

dtuite:
Prevents accidentally skipping tests when running tests in GitHub actions.


dtuite:
https://app.shortcut.com/larder/story/29752

PR #1660, opened 2025-10-26
Don't show terraform plugin in list

dtuite:


dtuite:
https://app.shortcut.com/larder/story/29726/add-backstage-terraform-provider-to-plugins-directory

